PR #13, opened 2022-05-10
migrate to kubeslice dns

kmjayadeep:
Refactor operator code to use kubeslice DNS. Removed old meshdns code


internal/dns/dns.go
```diff
@@ -1,143 +0,0 @@
-/*
```
bharath-avesha, line 1: Should we remove the dns directory as well?

kmjayadeep: it is removed

bharath-avesha (approved):

PR #21, opened 2022-05-13
feat(AM-2751): network policy integration 

rahulsawra98:


controllers/slice/slice.go
```diff
@@ -0,0 +1,507 @@
...
+			log.Error(err, "Failed to label namespace", "namespace", cfgAllowedNs)
+			return err
+		}
+		log.Info("Labeled namespace successfully", "namespace", cfgAllowedNs)
+
+		labeledAllowedNsList = append(labeledAllowedNsList, cfgAllowedNs)
+		statusChanged = true
+	}
+	// Sweep the existing namespaces again to unbind any namespace that was not found in the configured list
+	for existingAllowedNs, _ := range existingAllowedNsMap {
```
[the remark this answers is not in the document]
bharath-avesha: If someone removes from the list, the network policy gets updated. If you remove a ns called nginx-ingress from the allowed ns list, the network policy will remove the label kubeslice.io/namespace: nginx-ingress.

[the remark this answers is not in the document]
rahulsawra98: yaa but the label on the namespace still remains, this func just removes the label(kubeslice.io/namespace=.) from that ns

controllers/slice/slice.go
```diff
@@ -0,0 +1,507 @@
...
+		}
+		err := r.Delete(ctx, netPolicy)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				continue
+			}
+			log.Error(err, "NS unbind: Failed to remove slice netpol", "namespace", ns)
+		}
+	}
+	slice.Status.NetworkPoliciesInstalled = false
```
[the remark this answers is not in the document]
rahulsawra98: this is needed , if someone toggles off the namespaceIsolationEnabled flag , this flag indicates if the netpol are installed on cluster, and it uninstalls them.. https://github.com/kubeslice/worker-operator/pull/21/files#diff-729a2d9efde8ecdc3e0b52438f2e57dfe023b94457d0f42f9f986b971790e8cfR390

controllers/slice/slice.go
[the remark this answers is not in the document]
bharath-avesha: Then how will this work if the namespace is listed in allowed ns lists of more than one slice? If you remove the label here then other slices would break.

controllers/slice/slice.go
```diff
@@ -0,0 +1,507 @@
...
+
+func (r *SliceReconciler) cleanupSliceNamespaces(ctx context.Context, slice *kubeslicev1beta1.Slice) {
+	log := logger.FromContext(ctx).WithValues("type", "appNamespaces")
+
+	// Get the list of existing namespaces that are tagged with the kubeslice label
+	existingAppNsList := &corev1.NamespaceList{}
+	listOpts := []client.ListOption{
+		client.MatchingLabels(map[string]string{ApplicationNamespaceSelectorLabelKey: slice.Name}),
+	}
+	err := r.List(ctx, existingAppNsList, listOpts...)
```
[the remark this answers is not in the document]
rahulsawra98: I think this func will only be called at slice uninstallation, so not a heavy load on server!

internal/webhook/deploy/webhook.go
```diff
@@ -84,36 +94,87 @@ func Mutate(deploy *appsv1.Deployment, sliceName string) *appsv1.Deployment {
...
-func MutationRequired(metadata metav1.ObjectMeta) bool {
+func (wh *WebhookServer) MutationRequired(metadata metav1.ObjectMeta) (bool, string) {
 	annotations := metadata.GetAnnotations()
-
-	// no annotations available for the object, skip mutation
-	if annotations == nil {
-		return false
+	//early exit if metadata in nil
+	//we allow empty annotation, but namespace should not be empty
+	if len(annotations) == 0 && metadata.GetNamespace() == "" {
```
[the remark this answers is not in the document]
rahulsawra98: as per the comment from previous repo, it says "we allow empty annotation, but namespace should not be empty" ? so do you want me to remove len(annotations) == 0 check?

internal/webhook/deploy/webhook.go
```diff
@@ -84,36 +94,87 @@ func Mutate(deploy *appsv1.Deployment, sliceName string) *appsv1.Deployment {
...
+		}
+		// Check if namespace isolation is enabled for this slice
+		nsIsolationEnabled, err := wh.WhClient.GetSliceNamespaceIsolationPolicy(context.Background(), sliceNameInAnnotation)
+		if err != nil {
+			log.Error(err, "Could not get namespace isolation policy", "slice", sliceNameInAnnotation)
+			return false, ""
+		}
+		//if nsIsolationEnabled is not enabled at this point in time,we inject
+		if !nsIsolationEnabled {
+			return true, sliceNameInAnnotation
```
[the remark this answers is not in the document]
rahulsawra98: @bharath-avesha so what should be the logic for mutation required function?

api/v1beta1/slice_types.go
```diff
@@ -65,6 +65,18 @@ type SliceConfig struct {
 	SliceIpam SliceIpamConfig `json:"sliceIpam"`
 	// ExternalGatewayConfig determines istio ingress/egress configuration
 	ExternalGatewayConfig *ExternalGatewayConfig `json:"externalGatewayConfig,omitempty"`
+	// Namespace Isolation profile contains fields related to namespace binding to slice
+	NamespaceIsolationProfile NamespaceIsolationProfile `json:"namespaceIsolationProfile,omitempty"`
+}
```
kmjayadeep, line 70: All the optional fields should be pointers (those with `omitempty` flag). Otherewise, if you apply an object without defining these optional fields, they will be populated with golang default values. Also make sure there are nil checks everywhere

api/v1beta1/slice_types.go
```diff
@@ -65,6 +65,18 @@ type SliceConfig struct {
...
+
+// NamespaceIsolationProfile defines the namespace isolation policy for the slice
+type NamespaceIsolationProfile struct {
+	// Enable Namespace Isolation in the slice
+	IsolationEnabled bool `json:"isolationEnabled"`
```
kmjayadeep, line 75: I think this should be optional too. We can consider it to be disabled by default if not defined

internal/networkpolicy/reconciler.go
```diff
@@ -0,0 +1,255 @@
+package networkpolicy
```
kmjayadeep, line 1: missing copyright headers

internal/networkpolicy/reconciler.go
```diff
@@ -0,0 +1,255 @@
...
+
```
kmjayadeep, line 2: we are keeping all the reconcilers inside `controllers` folder under different packages. I think we should move this there to keep it consistent

kmjayadeep (changes requested):

internal/networkpolicy/reconciler.go
rahulsawra98: as per @bharath-avesha controllers which are not handling CRDs should be in internal/

api/v1beta1/slice_types.go
rahulsawra98: done

controllers/slice/namespaces.go
```diff
@@ -0,0 +1,497 @@
...
+		log.Error(err, "NS unbind: Failed to find namespace", "namespace", appNs)
+		return err
+	}
+
+	nsLabels := namespace.ObjectMeta.GetLabels()
+	_, ok := nsLabels[ApplicationNamespaceSelectorLabelKey]
+	if !ok {
+		debuglog.Info("NS unbind: slice label not found", "namespace", appNs)
+	} else {
+		// TBD: For now, we are just deleting the Avesha label from the namespace resource so that it becomes
```
bharath-avesha, line 259: Please remove this TBD since this would be taken care of by the annotation removal code from app deployment specs.

internal/hub/hubclient/hubclient.go
```diff
@@ -58,6 +58,7 @@ type HubClientRpc interface {
 	UpdateServiceExport(ctx context.Context, serviceexport *kubeslicev1beta1.ServiceExport) error
 	UpdateServiceExportEndpointForIngressGw(ctx context.Context, serviceexport *kubeslicev1beta1.ServiceExport,
 		ep *kubeslicev1beta1.ServicePod) error
+	UpdateAppNamesapces(ctx context.Context, sliceConfigName string, onboardedNamespaces []string) error
```
bharath-avesha, line 61: Typo in "Namesapces"

internal/webhook/deploy/webhook_utils.go
```diff
@@ -0,0 +1,32 @@
...
+type webhookClient struct {
+}
+
+func NewWebhookClient() *webhookClient {
+	return &webhookClient{}
+}
+
+func (w *webhookClient) SliceAppNamespaceConfigured(ctx context.Context, slice string, namespace string) (bool, error) {
+	return controllers.SliceAppNamespaceConfigured(ctx, slice, namespace)
+}
```
bharath-avesha, line 21: Please add newline before a new func definition starts.

PR #35, opened 2022-05-18
grpc stubs for UTs

kon3m:


internal/netop/netop.go
```diff
@@ -16,14 +16,32 @@
  *  limitations under the License.
  */
 
+/*
+ *  Copyright (c) 2022 Avesha, Inc. All rights reserved.
```
kmjayadeep, line 20: licence is already there at top. this is duplicate

internal/gwsidecar/gwsidecar.go
```diff
@@ -51,16 +51,27 @@ type GwConnectionContext struct {
 	RemoteSliceGwNsmSubnet string
 }
 
+type WorkerClientRpc interface {
```
kmjayadeep, line 54: where is this used?

internal/gwsidecar/gwsidecar.go
```diff
@@ -51,16 +51,27 @@ type GwConnectionContext struct {
...
+	GetStatus(ctx context.Context, serverAddr string) (*GwStatus, error)
+}
+
+type workerClientConfig struct {
+	Client sidecar.GwSidecarServiceClient
```
kmjayadeep, line 59: Can we rename `workerClientConfig` to something else which implies that this is a client to communicate with gw sidecar?

internal/gwsidecar/gwsidecar.go
```diff
@@ -51,16 +51,27 @@ type GwConnectionContext struct {
...
 	conn, err := grpc.Dial(serverAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	if err != nil {
 		return nil, err
 	}
 	defer conn.Close()
-
-	client := sidecar.NewGwSidecarServiceClient(conn)
-	res, err := client.GetStatus(ctx, &empty.Empty{})
+	worker.Client = sidecar.NewGwSidecarServiceClient(conn)
+	res, err := worker.Client.GetStatus(ctx, &empty.Empty{})
```
kmjayadeep, line 74: why are we storing the `Client` in `worker` struct, if it not used outside this scope? 
Should we run `sidecar.NewGwSidecarServiceClient(conn)` at the time of creating the struct and reuse the client everywhere?

internal/router/router.go
```diff
@@ -33,24 +33,30 @@ type SliceRouterConnCtx struct {
 	LocalNsmGwPeerIP       string
 }
 
-func GetClientConnectionInfo(ctx context.Context, addr string) ([]kubeslicev1beta1.AppPod, error) {
+type workerClientConfig struct {
+	client sidecar.SliceRouterSidecarServiceClient
```
kmjayadeep, line 37: same comment as above, naming is confusing

internal/router/router.go
```diff
@@ -22,7 +22,7 @@ import (
 	"context"
 
 	sidecar "github.com/kubeslice/router-sidecar/pkg/sidecar/sidecarpb"
-	kubeslicev1beta1 "github.com/kubeslice/worker-operator/api/v1beta1"
+	meshv1beta1 "github.com/kubeslice/worker-operator/api/v1beta1"
 	"google.golang.org/grpc"
```
kmjayadeep, line 26: why do we rename `kubeslicev1beta1` to `meshv1beta1`?

kmjayadeep (changes requested):

internal/gwsidecar/gwsidecar.go
kon3m: @YachikaRalhan

internal/router/router.go
kon3m: changed it

internal/gwsidecar/gwsidecar.go
kon3m: done

internal/router/router.go
kon3m: Done

internal/netop/netop.go
kon3m: removed it

internal/gwsidecar/gwsidecar.go
YachikaRalhan: The problem is client type is different for the all components like a gateway, router, etc. So that is why each component has it's own interface and own client type.

kmjayadeep: My question is why are we assigning grpc client to `worker.Client` variable?
This variable is not used anywhere outside this function scope.

PR #36, opened 2022-05-20
feat(AM-4013) New Namespace Reconciler to update cluster CR

YachikaRalhan:
Add Reconciler to Monitor the Namespaces create, delete and update to the Cluster CR along with the Slice Name


kmjayadeep:
please add a description and proper title to the PR

config/rbac/role.yaml
```diff
@@ -39,6 +39,18 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - create
```
bharath-avesha, line 47: Do we need create, delete? I think we would just need get, list and watch for this feature.

internal/hub/hubclient/hubclient.go
```diff
@@ -146,6 +147,132 @@ func updateClusterInfoToHub(ctx context.Context, spokeclient client.Client, hubC
 	return nil
 }
 
+func UpdateNamespaceInfoToHub(ctx context.Context, hubClient client.Client, onboardNamespace, sliceName string) error {
+	hubCluster := &hubv1alpha1.Cluster{}
+	err := hubClient.Get(ctx, types.NamespacedName{
+		Name:      os.Getenv("CLUSTER_NAME"),
+		Namespace: os.Getenv("HUB_PROJECT_NAMESPACE"),
```
bharath-avesha, line 154: Please reuse ProjectNamespace global var here.

internal/hub/hubclient/hubclient.go
```diff
@@ -146,6 +147,132 @@ func updateClusterInfoToHub(ctx context.Context, spokeclient client.Client, hubC
 	return nil
 }
 
+func UpdateNamespaceInfoToHub(ctx context.Context, hubClient client.Client, onboardNamespace, sliceName string) error {
+	hubCluster := &hubv1alpha1.Cluster{}
+	err := hubClient.Get(ctx, types.NamespacedName{
+		Name:      os.Getenv("CLUSTER_NAME"),
```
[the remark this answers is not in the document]
YachikaRalhan: Same explanation here also @bharath-avesha 
`Yes you are I could have used but it test cases it is behaving weirdly, as in it not getting the global variable value here and in result test cases are failing`

internal/hub/hubclient/hubclient.go
```diff
@@ -146,6 +147,132 @@ func updateClusterInfoToHub(ctx context.Context, spokeclient client.Client, hubC
...
+		hubCluster.Status.Namespaces = append(hubCluster.Status.Namespaces, hubv1alpha1.NamespacesConfig{
+			Name:      onboardNamespace,
+			SliceName: sliceName,
+		})
+		log.Info("Adding namespace on hub cluster", "cluster", ClusterName, "namespace", onboardNamespace)
+		return updateNamespaceInfoToHub(hubClient, ctx, hubCluster, onboardNamespace, sliceName)
+	}
+	return nil
+}
+func updateNamespaceInfoToHub(hubClient client.Client, ctx context.Context,
```
bharath-avesha, line 178: Add a new line before the func definition.

internal/hub/hubclient/hubclient.go
```diff
@@ -146,6 +147,132 @@ func updateClusterInfoToHub(ctx context.Context, spokeclient client.Client, hubC
...
+	log.Info("Deleting namespace on hub cluster", "cluster", ClusterName, "namespace", onboardNamespace)
+	return deleteNamespaceInfoToHub(hubClient, ctx, hubCluster, onboardNamespace)
+}
+
+func deleteNamespaceInfoToHub(hubClient client.Client, ctx context.Context,
+	hubCluster *hubv1alpha1.Cluster, onboardNamespace string) error {
+	if err := hubClient.Status().Update(ctx, hubCluster); err != nil {
+		if errors.IsConflict(err) {
+			log.Error(err, "conflict error", "namespace", onboardNamespace)
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
bharath-avesha, line 240: Can we use retry.RetryOnConflict() directly instead of using it only if the first update attempt fails? It would look much cleaner that way.

internal/namespace/controllers/reconciler.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (c *Reconciler) getSliceNameFromNs(ns string) (string, error) {
+	namespace := corev1.Namespace{}
+	err := c.Client.Get(context.Background(), types.NamespacedName{Name: ns}, &namespace)
+	if err != nil {
+		c.Log.Error(err, "error while retrieving namespace")
+		return "", err
+	}
+	// TODO: label should not be hardcoded, it should come as a constant from slice reconciler pkg
+	return namespace.Labels["kubeslice.io/slice"], nil
```
bharath-avesha, line 40: this will crash if the ns does not have any labels on it, dereferencing a null map.

bharath-avesha: If the labels are present please check if the slice label is present. test for presence:

sliceName, present := labels[""]
if !present {
   return "", nil
}

return sliceName, nil

YachikaRalhan: No no it won't crash, it will just return an empty string. I have tested that scenario as well

internal/hub/hubclient/hubclient.go
YachikaRalhan: Sure let me update it

internal/namespace/controllers/reconciler.go
YachikaRalhan: Please see this example - https://go.dev/play/p/W7-bqAImRLc
even if it is present or not we will be adding an empty string only, so we don't need to check whether the key is present or not in map

YachikaRalhan: I have added a new commit to handle the case when namespace labels are nil.

rahulsawra98 (approved):
LGTM

PR #41, opened 2022-05-27
Fix(AM-4125)- retry updating node ip in cluster CR

YachikaRalhan:


internal/hub/hubclient/hubclient.go
```diff
@@ -110,38 +110,44 @@ func PostClusterInfoToHub(ctx context.Context, spokeclient client.Client, hubCli
...
+		cniSubnet, err := c.GetNsmExcludedPrefix(ctx, "nsm-config", "kubeslice-system")
+		if err != nil {
+			log.Error(err, "Error getting cni Subnet")
+			return err
+		}
+		log.Info("cniSubnet", "cniSubnet", cniSubnet)
+		hubCluster.Spec.ClusterProperty.GeoLocation.CloudRegion = clusterInfo.ClusterProperty.GeoLocation.CloudRegion
+		hubCluster.Spec.ClusterProperty.GeoLocation.CloudProvider = clusterInfo.ClusterProperty.GeoLocation.CloudProvider
+		hubCluster.Spec.NodeIP = nodeIP
+		hubCluster.Status.CniSubnet = cniSubnet
```
rahulsawra98, line 138: this is not required here!

internal/hub/hubclient/hubclient.go
```diff
@@ -110,38 +110,43 @@ func PostClusterInfoToHub(ctx context.Context, spokeclient client.Client, hubCli
...
-	if err := hubClient.Update(ctx, hubCluster); err != nil {
-		log.Error(err, "Error updating to cluster spec on hub cluster")
-		return err
-	}
-	hubCluster.Status.CniSubnet = cniSubnet
-	if err := hubClient.Status().Update(ctx, hubCluster); err != nil {
-		log.Error(err, "Error updating cniSubnet to cluster status on hub cluster")
 		return err
 	}
 	return nil
```
kmjayadeep, line 152:
```
if err != nil {
   return err
}
return nil
```

is same as 
```
return err
```

rahulsawra98 (approved):
LGTM! Thanks for quick PR

PR #48, opened 2022-06-08
Handle nsm IP address change of slice gw and app pods

bharath-avesha:
Handled the case of nsm IP address change of slice gw due to nsm control
plane or data plane pods restarting.

For Istio, handled the case of nsm IP address change of an app pod that is
exporting a service. Updated the corresponding Istio ServiceEntry
object.


rahulsawra98 (approved):
LGTM!

gourishkb (approved):

PR #53, opened 2022-06-13
fix(AM-AddNodePortToGWStatus): starts a NodePort service on the same …

kon3m:
…port as the previous one

Signed-off-by: kon3m <aakash.dokka@aveshasystems.com>

### DIFF
--- a/controllers/slicegateway/slicegateway.go
+++ b/controllers/slicegateway/slicegateway.go
@@ -323,7 +323,7 @@ func (r *SliceGwReconciler) serviceForGateway(g *kubeslicev1beta1.SliceGateway) 
 			}},
 		},
 	}
-	if g.Status.Config.SliceGatewayNodePort >= 30000 && g.Status.Config.SliceGatewayNodePort <= 33000 {
+	if g.Status.Config.SliceGatewayNodePort != 0 {
 		svc.Spec.Ports[0].NodePort = int32(g.Status.Config.SliceGatewayNodePort)
 	}
 	ctrl.SetControllerReference(g, svc, r.Scheme)
--- a/controllers/slicegateway/slicegateway.go
+++ b/controllers/slicegateway/slicegateway.go
@@ -305,23 +305,46 @@ func (r *SliceGwReconciler) deploymentForGatewayServer(g *kubeslicev1beta1.Slice
 }
 
 func (r *SliceGwReconciler) serviceForGateway(g *kubeslicev1beta1.SliceGateway) *corev1.Service {
-	svc := &corev1.Service{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "svc-" + g.Name,
-			Namespace: g.Namespace,
-			Labels: map[string]string{
-				"kubeslice.io/slice": g.Spec.SliceName,
+	var svc *corev1.Service
+	if g.Status.Config.SliceGatewayNodePort == 0 {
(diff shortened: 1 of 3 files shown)


controllers/slicegateway/slicegateway.go
```diff
@@ -302,23 +302,46 @@ func (r *SliceGwReconciler) deploymentForGatewayServer(g *kubeslicev1beta1.Slice
...
+				Type:     "NodePort",
+				Selector: labelsForSliceGwDeployment(g.Name, g.Spec.SliceName),
+				Ports: []corev1.ServicePort{{
+					Port:       11194,
+					Protocol:   corev1.ProtocolUDP,
+					TargetPort: intstr.FromInt(11194),
+					NodePort:   int32(g.Status.Config.SliceGatewayNodePort),
+				}},
+			},
+		}
```
kmjayadeep, line 344: You don't need the `if` block here. The code in `else` block will cover both the cases.

If `g.Status.Config.SliceGatewayNodePort` is 0, then `NodePort` will be 0 and it will be automatically generated when applied.

internal/hub/controllers/slicegateway_controller.go
```diff
@@ -190,6 +190,7 @@ func (r *SliceGwReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		SliceGatewaySubnet:          sliceGw.Spec.LocalGatewayConfig.GatewaySubnet,
 		SliceGatewayRemoteSubnet:    sliceGw.Spec.RemoteGatewayConfig.GatewaySubnet,
 		SliceGatewayHostType:        sliceGw.Spec.GatewayHostType,
+		SliceGatewayNodePort:        sliceGw.Spec.LocalGatewayConfig.NodePort,
```
kmjayadeep, line 193: Do we already have this field in the CRD?

kmjayadeep (changes requested):

controllers/slicegateway/slicegateway.go
```diff
@@ -302,23 +302,46 @@ func (r *SliceGwReconciler) deploymentForGatewayServer(g *kubeslicev1beta1.Slice
...
+				Selector: labelsForSliceGwDeployment(g.Name, g.Spec.SliceName),
+				Ports: []corev1.ServicePort{{
+					Port:       11194,
+					Protocol:   corev1.ProtocolUDP,
+					TargetPort: intstr.FromInt(11194),
+				}},
+			},
+		}
+	} else {
+		svc = &corev1.Service{
```
bharath-avesha, line 326: @kon3m To avoid repetitive code, please change this to:

Leave the struct init code as is:
svc := &corev1.Service{
...
}

then add this (please double check the node port range):
if g.Status.Config.SliceGatewayNodePort >= 30000 && g.Status.Config.SliceGatewayNodePort <= 33000 {
    svc.Spec.Ports[0].NodePort = int32(g.Status.Config.SliceGatewayNodePort)
}

internal/hub/controllers/slicegateway_controller.go
kon3m: yes, we do : [crd](https://github.com/kubeslice/charts/blob/master/charts/kubeslice-worker/crds/networking.kubeslice.io_slicegateways.yaml#:~:text=sliceGatewayNodePort%3A)

bharath-avesha (approved):

PR #54, opened 2022-06-14
Move internal packages to pkg folder

kmjayadeep:
packages inside `internal` folder are moved to `pkg` folder so they can be imported in other repos


rahulsawra98 (approved):
Awesome!

YachikaRalhan (approved):

PR #55, opened 2022-06-14
feat(AM-4303): node restart

rahulsawra98:
Signed-off-by: Rahul <rahul.sawra@aveshasystems.com>


controllers/slicegateway/slicegateway.go
```diff
@@ -678,3 +680,69 @@ func (r *SliceGwReconciler) SyncNetOpConnectionContextAndQos(ctx context.Context
 	debugLog.Info("netop pods sync complete", "pods", r.NetOpPods)
 	return nil
 }
+
+func (r *SliceGwReconciler) createHeadlessServiceForGwServer(slicegateway *kubeslicev1beta1.SliceGateway) *corev1.Service {
+	svc := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      slicegateway.Status.Config.SliceGatewayRemoteGatewayID,
```
kmjayadeep, line 687: probably a good idea to prefix the service name with slice name. Otherwise when you run `k get svc` it is difficult to determine which service belongs to which slice

controllers/slicegateway/slicegateway.go
```diff
@@ -678,3 +680,69 @@ func (r *SliceGwReconciler) SyncNetOpConnectionContextAndQos(ctx context.Context
...
+}
+
+func (r *SliceGwReconciler) reconcileNodes(ctx context.Context, slicegateway *kubeslicev1beta1.SliceGateway) error {
+	log := r.Log
+	//TODO:fetch currentNodeIP from controller cluster CR?
+	currentNodeIP := r.NodeIP
+	nodeIpList := cluster.GetNodeExternalIpList()
+	if len(nodeIpList) == 0 {
+		//err := errors.New("node IP list is empty")
+		return nil
```
kmjayadeep, line 727: this should return an error and fail the reconciliation

internal/cluster/node.go
```diff
@@ -78,6 +85,10 @@ func (n *NodeInfo) populateNodeIpList() error {
 		return fmt.Errorf("can't fetch node list: %+v ", err)
 	}
 
+	if len(nodeList.Items) == 0 {
+		return fmt.Errorf("can't fetch node list: %+v ", err)
+	}
+
```
kmjayadeep, line 91: if none of the nodes have externalIps available, should we consider internalIp too? This would make it easier for on-prem and kind installations. For eg. in k3s installation, it shows node public ip as `internalIP`

This is from my kind node:

```
    addresses:
    - address: 172.19.0.2
      type: InternalIP
    - address: kind-control-plane
      type: Hostname
```
Here the IP is docker network ip, which is reachable from other kind clusters. It should be considered as NodeIP even through it is configured as `InternalIP` here.

What do you think?

internal/cluster/node_reconciler.go
```diff
@@ -0,0 +1,76 @@
+/*
```
kmjayadeep, line 1: we should avoid adding reconciler code in `internal` package. It should be in controllers package.

internal/cluster/node.go
```diff
@@ -91,15 +102,20 @@ func (n *NodeInfo) populateNodeIpList() error {
...
 	nodeIP := nodeIPs[0]
 	log.Info("nodeIP selected", "nodeIP ", nodeIP)
 	return nodeIP, err
 }
+
+func GetNodeExternalIpList() []string {
+	nodeInfo.Lock()
+	defer nodeInfo.Unlock()
+	return nodeInfo.ExternalIP
```
kmjayadeep, line 120: `nodeInfo.ExternalIP` should be `nodeInfo.ExternalIPList` or `nodeInfo.ExternalIPs` since it is a list of Ip addresses

internal/cluster/node.go
```diff
@@ -91,15 +102,20 @@ func (n *NodeInfo) populateNodeIpList() error {
...
 
 func GetNodeIP(client client.Client) (string, error) {
-	nodeInfo := &NodeInfo{
-		Client: client,
-	}
-	nodeIPs, err := nodeInfo.getNodeExternalIpList(client)
-	if err != nil {
+	nodeInfo.Client = client
+	nodeIPs, err := nodeInfo.getNodeExternalIpList()
+	if err != nil || len(nodeIPs) == 0 {
```
kmjayadeep, line 107: it is better to pass `client` as argument as in the previous code. Why is this changed?

main.go
```diff
@@ -90,7 +91,7 @@ func main() {
 	flag.Parse()
 
 	ctrl.SetLogger(logger.NewLogger())
-
+	fmt.Println("enableLeaderElection",enableLeaderElection)
```
kmjayadeep, line 94: can we remove this comment and the `fmt` import?

main.go
```diff
@@ -159,10 +169,18 @@ func main() {
...
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "SliceGw")
 		os.Exit(1)
 	}
+	if err := (&cluster.NodeReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("controllers").WithName("node reconciller"),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "node")
+		//os.Exit(1)
```
kmjayadeep, line 182: uncomment `os.Exit`

tests/spoke/slice_netpol_test.go
```diff
@@ -492,7 +492,7 @@ var _ = Describe("SliceNetpol", func() {
 					client.InNamespace(slice.Namespace),
 				}
 				err = k8sClient.List(ctx, &eventList, opts...)
-				return len(eventList.Items) > 0
+				return len(eventList.Items) > 1
```
kmjayadeep, line 495: is this change related to this PR?

tests/spoke/slicegw_controller_test.go
```diff
@@ -338,7 +338,8 @@ var _ = Describe("Worker SlicegwController", func() {
 			}, time.Second*10, time.Millisecond*250).Should(BeTrue())
 
 			createdSliceGw.Status.Config.SliceGatewayHostType = "Client"
-
+			createdSliceGw.Status.Config.SliceGatewayRemoteNodeIP = "1.2.3.4"
+			createdSliceGw.Status.Config.SliceGatewayRemoteGatewayID = "remote-gateway-id"
 			Eventually(func() bool {
```
kmjayadeep, line 343: Can you add test cases for the feature? Or do you want to do in a separate PR?

main.go
```diff
@@ -130,6 +131,15 @@ func main() {
 		setupLog.Error(err, "could not create spoke netop client for slice gateway reconciler")
 		os.Exit(1)
 	}
+
+	clientForHubMgr, err := client.New(ctrl.GetConfigOrDie(), client.Options{
+		Scheme: scheme,
+	})
+	//check if user has provided NODE_IP as env variable, if not fetch the ExternalIP from gateway nodes
+	nodeIP, err := cluster.GetNodeIP(clientForHubMgr)
```
kmjayadeep, line 139: where does it check `NODE_IP` env?

main.go
```diff
@@ -159,10 +169,18 @@ func main() {
 		WorkerRouterClient:    workerRouterClient,
 		WorkerNetOpClient:     workerNetOPClient,
 		EventRecorder:         sliceGwEventRecorder,
+		NodeIP:                nodeIP,
```
kmjayadeep, line 172: I don't think we need to pass the dynamic NodeIP from there. The same will be fetched by `r.reconcileNodes` method. Instead, we should have a separate variable for NodeIP from env variable.

something like:

```
ProvidedNodeIP: os.Env("NODE_IP")
```

in slicegatewayreconciler, it should check if this variable is set (which means Node IP is provided as env variable) and then ignore the reconciliation. You don't even have to start NodeReconciler if this is set.

kmjayadeep (changes requested):
Make sure both the cases are tested - where Node IP is passed as env variable and not passed as env variable

controllers/slicegateway/slicegateway.go
rahulsawra98: yeah, makes sense!

controllers/slicegateway/slicegateway.go
rahulsawra98: yeah, the test cases were failing , thats why for now I just returned nil, will mock this func

internal/cluster/node.go
rahulsawra98: sure, I will include this

internal/cluster/node.go
rahulsawra98: yeah

internal/cluster/node.go
rahulsawra98: nodeInfo struct already has a client as a field

tests/spoke/slicegw_controller_test.go
rahulsawra98: yeah I will add this as a part of this PR

main.go
rahulsawra98: https://github.com/kubeslice/worker-operator/blob/master/pkg/cluster/node.go#L51

main.go
rahulsawra98: but what if the provided nodeIP is no longer available? do we not want to select another nodeIP in that case?

internal/cluster/node.go
rahulsawra98: Will do this in separate story

pkg/cluster/node.go
```diff
@@ -108,7 +108,6 @@ func GetNodeIP(client client.Client) (string, error) {
 		log.Error(err, "Getting NodeIP From kube-api-server")
 		os.Exit(1)
 	}
-
 	nodeIP := nodeIPs[0]
```
gourishkb, line 111: we can add for len(nodeIPs) > 0 before doing this

rahulsawra98: @gourishkb added similar check on line 107,checking if len is 0

controllers/slicegateway/reconciler.go
```diff
@@ -192,8 +192,75 @@ func (r *SliceGwReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		//reconcile the available nodes
+		if err := r.reconcileNodes(ctx, sliceGw); err != nil {
+			return ctrl.Result{}, err
+		}
 	}
 
+	isClient := sliceGw.Status.Config.SliceGatewayHostType == "Client"
+	if isClient {
+		//check if SliceGatewayRemoteGatewayID is populated , else requeue
+		if sliceGw.Status.Config.SliceGatewayRemoteNodeIP == "" || sliceGw.Status.Config.SliceGatewayRemoteGatewayID == "" {
```
bharath-avesha, line 204: This check should be avoided, instead readyToDeployGwClient() should be enhanced.

controllers/slicegateway/reconciler.go
```diff
@@ -192,8 +192,75 @@ func (r *SliceGwReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+				}
+				log.Info("Created a new Endpoint for sliceGw", "sliceGw", sliceGw.Name)
+				return ctrl.Result{Requeue: true}, nil
+			} else {
+				log.Error(err, "Failed to get an endpoint")
+				return ctrl.Result{}, err
+			}
+		}
+		debugLog.Info("SliceGatewayRemoteNodeIP", "SliceGatewayRemoteNodeIP", sliceGw.Status.Config.SliceGatewayRemoteNodeIP)
+		debugLog.Info("from endpointFound", "from endpointFound", endpointFound.Subsets[0].Addresses[0].IP)
```
bharath-avesha, line 251: This log can be moved inside the 'if' block on line 253. The log on line 254 can print the value in endpointFound.

controllers/slicegateway/reconciler.go
```diff
@@ -192,8 +192,75 @@ func (r *SliceGwReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 			return ctrl.Result{}, err
 		}
+		//reconcile the available nodes
+		if err := r.reconcileNodes(ctx, sliceGw); err != nil {
+			return ctrl.Result{}, err
+		}
 	}
 
+	isClient := sliceGw.Status.Config.SliceGatewayHostType == "Client"
+	if isClient {
```
bharath-avesha, line 202: This whole block can be a separate func. r.ReconcileGwHeadlessSvc()

controllers/slicegateway/types.go
```diff
@@ -36,6 +36,8 @@ type NetOpPod struct {
 
 type HubClientProvider interface {
 	UpdateNodePortForSliceGwServer(ctx context.Context, sliceGwNodePort int32, sliceGwName string) error
+	UpdateNodeIpInCluster(ctx context.Context, clusterName, nodeIP, namespace string) error
```
bharath-avesha, line 39: The cluster name and namespace should not matter to the slice gw. It should be restricted to the hub client pkg. This func should only take ctx and nodeIP as input.

controllers/slicegateway/types.go
```diff
@@ -36,6 +36,8 @@ type NetOpPod struct {
...
+	GetClusterNodeIP(ctx context.Context, clusterName, namespace string) (string, error)
```
bharath-avesha, line 40: Should only have one param ctx.

pkg/cluster/node.go
```diff
@@ -78,28 +85,36 @@ func (n *NodeInfo) populateNodeIpList() error {
...
-	nodeInfo := &NodeInfo{
-		Client: client,
-	}
-	nodeIPs, err := nodeInfo.getNodeExternalIpList(client)
-	if err != nil {
+	nodeInfo.Client = client
+	nodeIPs, err := nodeInfo.getNodeExternalIpList()
+	if err != nil || len(nodeIPs) == 0 {
 		log.Error(err, "Getting NodeIP From kube-api-server")
 		os.Exit(1)
```
bharath-avesha, line 109: Why are we exiting the proc here? The caller of the func should decide if the proc should exit or not.

pkg/cluster/node_reconciler.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, _x := range x {
+		// 0 value for int is 0, so just increment a counter for the string
+		diff[_x]++
+	}
+	for _, _y := range y {
+		// If the string _y is not in diff bail out early
+		if _, ok := diff[_y]; !ok {
+			return false
+		}
+		diff[_y] -= 1
```
bharath-avesha, line 93: Not sure if i see why lines 93-96 are needed. After this for loop ends, cant we simply return true?

controllers/slicegateway/reconciler.go
```diff
@@ -192,8 +192,75 @@ func (r *SliceGwReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			)
 			return ctrl.Result{}, err
 		}
+		//reconcile the available nodes
+		if err := r.reconcileNodes(ctx, sliceGw); err != nil {
+			return ctrl.Result{}, err
+		}
 	}
 
+	isClient := sliceGw.Status.Config.SliceGatewayHostType == "Client"
```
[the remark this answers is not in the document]
rahulsawra98: sure!

controllers/slicegateway/reconciler.go
rahulsawra98: yeah that would make code clean

pkg/cluster/node.go
rahulsawra98: sure, I will make changes

PR #60, opened 2022-06-23
fix(AM-4381): add retry func while updating namespaces in workersliceconfig

rahulsawra98:
when the onboarded namespaces are updated to workersliceconfig , if the update fails due to conflict error , the user will not be able to delete sliceconfig! 
This PR wraps update logic under retry func


gourishkb (approved):
LGTM

YachikaRalhan (approved):

PR #63, opened 2022-06-28
integration tests for node restart feature

rahulsawra98:
This PR is in continuation of #55.
Integration Tests for node restart feature

### DIFF
--- a/tests/spoke/cluster_info_test.go
+++ b/tests/spoke/cluster_info_test.go
@@ -28,6 +28,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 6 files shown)


tests/spoke/cluster_info_test.go
```diff
@@ -78,13 +79,24 @@ prefixes:
 - 192.168.0.0/16
 - 10.96.0.0/12`)
 
+			DeferCleanup(func() {
+				Expect(k8sClient.Delete(ctx, node)).Should(Succeed())
```
Rahul-D78, line 83: Can we wait until the object is completely deleted .

Rahul-D78 (commented):
LGTM :+1:

tests/spoke/cluster_info_test.go
rahulsawra98: yeah we can , but we are not fetching the node again. We just want to trigger the node reconciler

Rahul-D78 (approved):

PR #64, opened 2022-07-01
Feature e2e

saivenkatesh1909:
The above changes will build worker-operator and will perform a e2e test on every pull request. currently this works on push, but we will modify the script to use the on push.

Also we need to the change the secret/personal access token and also the e2e repo to complete the function. Please verify


.github/workflows/scripts/kind-nodeip.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Get the Host IP
+HOST_IP=$(hostname -i)
+echo $HOST_IP
+
+sed -i "s/<NEW_IP>/${HOST_IP}/g" .github/workflows/scripts/kind-controller.yaml
```
kmjayadeep, line 7: shouldn't you be using node ip of the kind cluster instead of host ip?

.github/workflows/end-to-end-test.yaml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+    - name: checkout the repo
+      uses: actions/checkout@v3
+    - name: build the worker-operator
+      uses: docker/build-push-action@v2
+      with:
+        tags: worker-operator:latest
+        push: false
+    - name: determine node-ip
+      run:  sh .github/workflows/scripts/kind-nodeip.sh
```
kmjayadeep, line 15: `determine node-ip` step should be after creating the kind clusters. It is not possible to get node ip before right?

.github/workflows/end-to-end-test.yaml
```diff
@@ -0,0 +1,55 @@
...
+      uses: docker/build-push-action@v2
+      with:
+        context: ./kubeslice-e2e
+        file: kubeslice-e2e/test-prod.Dockerfile
+        build-args: PROFILE=profile_kind_two
+        tags: end-to-end:latest
+        push: false
+    - name: Docker Extract
+      uses: shrink/actions-docker-extract@v1
+      id: extract
```
kmjayadeep, line 40: `actions-docker-extract` is to extract files from an image. Here we need to do `docker run` with mounted volumes instead

.github/workflows/end-to-end-test.yaml
```diff
@@ -0,0 +1,55 @@
...
+      uses: docker/build-push-action@v2
+      with:
+        tags: worker-operator:latest
+        push: false
+    - name: determine node-ip
+      run:  sh .github/workflows/scripts/kind-nodeip.sh
+    - name: create kind clusters
+      run: sh .github/workflows/scripts/kind.sh
+    - name: clone the e2e repo
+      uses: actions/checkout@v3
```
kmjayadeep, line 19: No need to clone the repo and build it here. the docker image can be built in a separate pipeline for e2e repo and push to docker registry. We can just pull the image and use it

.github/workflows/scripts/kind-worker-1.yaml
```diff
@@ -0,0 +1,19 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  # WARNING: It is _strongly_ recommended that you keep this the default
+  # (127.0.0.1) for security reasons. However it is possible to change this.
+  apiServerAddress: "<NEW_IP>"
```
kmjayadeep, line 6: why do we need to expose it in the host ip address?

kmjayadeep (changes requested):

saivenkatesh1909 (commented):
updated the changes with the new pull request

kmjayadeep (approved):

PR #93, opened 2022-08-29
fix(AM-4902): fix nodePort value populating as 0

rahulsawra98:
While refactoring changes for sonarqube warnings , sliceGwNodePort was passed to a different function but its value was never returned back hence the value was being set as 0.
This PR fixes this issue

### DIFF
--- a/controllers/slicegateway/reconciler.go
+++ b/controllers/slicegateway/reconciler.go
@@ -144,7 +144,7 @@ func (r *SliceGwReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// true if the gateway is openvpn server
 	// Check if the Gw service already exists, if not create a new one if it is a server
 	if isServer(sliceGw) {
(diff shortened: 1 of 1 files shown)


controllers/slicegateway/reconciler.go
```diff
@@ -144,7 +144,7 @@ func (r *SliceGwReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// true if the gateway is openvpn server
 	// Check if the Gw service already exists, if not create a new one if it is a server
 	if isServer(sliceGw) {
-		reconcile, result, err = r.handleSliceGwSvcCreation(ctx, sliceGw, sliceGwNodePort)
+		reconcile, result, sliceGwNodePort ,err = r.handleSliceGwSvcCreation(ctx, sliceGw)
```
bharath-avesha, line 147: Please fix indentation around the ','

reconcile, result, sliceGwNodePort, err

Applies to the rest of the diff as well. There should be a space between the comma and the next var, but not between a var and a comma.

Please run go fmt.

rahulsawra98: done

PR #96, opened 2022-09-01
Fix webhook to support non deployment workloads

Rahul-D78:
This fix should have webhook to support non deployment workloads

### DIFF
--- a/controllers/slice/reconciler.go
+++ b/controllers/slice/reconciler.go
@@ -69,7 +69,7 @@ type SliceReconciler struct {
 //+kubebuilder:rbac:groups=networking.istio.io,resources=gateways,verbs=get;list;create;update;watch;delete
 //+kubebuilder:rbac:groups=networking.istio.io,resources=serviceentries,verbs=get;list;create;update;watch;delete
 //+kubebuilder:rbac:groups=networking.istio.io,resources=virtualservices,verbs=get;list;create;update;watch;delete
-//+kubebuilder:webhook:path=/mutate-appsv1-deploy,mutating=true,failurePolicy=fail,groups="apps",resources=deployments,verbs=create;update,versions=v1,name=mdeploy.kubeslice.io,admissionReviewVersions=v1,sideEffects=NoneOnDryRun
+//+kubebuilder:webhook:path=/mutate-appsv1-deploy,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mdeploy.kubeslice.io,admissionReviewVersions=v1,sideEffects=NoneOnDryRun
 //+kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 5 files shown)


controllers/slice/reconciler.go
```diff
@@ -69,7 +69,7 @@ type SliceReconciler struct {
 //+kubebuilder:rbac:groups=networking.istio.io,resources=gateways,verbs=get;list;create;update;watch;delete
 //+kubebuilder:rbac:groups=networking.istio.io,resources=serviceentries,verbs=get;list;create;update;watch;delete
 //+kubebuilder:rbac:groups=networking.istio.io,resources=virtualservices,verbs=get;list;create;update;watch;delete
-//+kubebuilder:webhook:path=/mutate-appsv1-deploy,mutating=true,failurePolicy=fail,groups="apps",resources=deployments,verbs=create;update,versions=v1,name=mdeploy.kubeslice.io,admissionReviewVersions=v1,sideEffects=NoneOnDryRun
+//+kubebuilder:webhook:path=/mutate-appsv1-deploy,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mdeploy.kubeslice.io,admissionReviewVersions=v1,sideEffects=NoneOnDryRun
```
kmjayadeep, line 72: should we change the service path and webhook name as well?  it looks a bit weird to call the webook `mdeploy` when it is actually mutating pod instead of deployments.
This will require helm chart changes though

kmjayadeep: helm chart has to be changed anyway. so i would say lets fix the names as well

Rahul-D78: can we call it webhook.kubeslice.io

rahulsawra98: yeah that sounds good @Rahul-D78

pkg/webhook/deploy/webhook.go
```diff
@@ -56,22 +57,26 @@ type WebhookServer struct {
...
+	pod := &corev1.Pod{}
+	err := wh.decoder.Decode(req, pod)
 	if err != nil {
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 	log := logger.FromContext(ctx)
 
-	if mutate, sliceName := wh.MutationRequired(deploy.ObjectMeta); !mutate {
-		log.Info("mutation not required", "pod metadata", deploy.Spec.Template.ObjectMeta)
+	if pod.ObjectMeta.Namespace == "" {
```
rahulsawra98, line 67: can you add a comment here , explaining why we re doing this?

rahulsawra98:
@Rahul-D78 please also change the mutatingwebhookconfiguration https://github.com/kubeslice/dev-charts/blob/master/charts/kubeslice-worker/templates/webhook.yaml

bharath-avesha (approved):

kmjayadeep (approved):

rahulsawra98:
hey @Rahul-D78 please squash the commits before you merge. Also raise a PR for helm charts
Thanks for the PR

rahulsawra98 (approved):

PR #126, opened 2022-10-27
add a new label to application ns and kubeslice-system ns to make nsm webhook fault tolerant

rahulsawra98:
This PR adds a new label kubeslice.io/inject=true to application namespaces and specifically to kubeslice-system namespace after a slice is created.
We will use this label in nsm mutatingwebhookconfig in namespace selector. This way only requests from application namespaces and kubeslice-system(gw pods) goes to nsm webhook

### DIFF
--- a/controllers/slice/utils.go
+++ b/controllers/slice/utils.go
@@ -23,7 +23,9 @@ import (
 
 	kubeslicev1beta1 "github.com/kubeslice/worker-operator/api/v1beta1"
 	"github.com/kubeslice/worker-operator/controllers"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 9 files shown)


NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2022-11-07T08:13:38-master-214/index.html'

controllers/slice/utils.go
```diff
@@ -35,6 +37,40 @@ func (r *SliceReconciler) cleanupSliceResources(ctx context.Context, slice *kube
 	r.cleanupSliceRouter(ctx, slice.Name)
 	//cleanup Service Discovery objects - serviceimport and export objects that belong to this slice
 	r.cleanupServiceDiscoveryObjects(ctx, slice.Name)
+	// remove kubeslice.io/inject label from kubeslice-system , in case this is last slice
+	r.removeLabel(ctx)
+}
+
+// this func removes the kubeslice.io/inject label from kubeslice-system , once the last slice has been deleted
+func (r *SliceReconciler) removeLabel(ctx context.Context) error {
```
bharath-avesha, line 45: Should we even do this? The label should remain as long as the kubeslice-system namespace is present. I dont think it should be predicated on the presence of slices.

rahulsawra98: @bharath-avesha this is done in case someone uninstalls kubeslice-worker and if the label is present on kubeslice-system, the next installation will cause troubles! Hence, after the last slice is uninstalled,I removed the label.
In case , the slice is again installed, it bring backs the label again.

bharath-avesha: Hmmm,  I am not sure why it would cause problems in the next installation if the ns is already labeled. So if someone manually adds this label for fun before installing then our installation will fail? If something fails in this func, we would unnecessarily propagate an error and be in an error state. So if avoidable, we should get rid of this func.

bharath-avesha (approved):

controllers/slice/utils.go
rahulsawra98: @bharath-avesha ,lets say If the label remains on kubeslice-system namespace even after uninstalling worker-operator.
When the user performs next installation (the label is still present), when the other pods are admitted , the requests will go to webhookconfig , but the nsm webhook server is still not present. Hence the pods will not be admitted and be in error state continously
I hope that answers your doubt?

PR #127, opened 2022-11-03
feat(  AM-5594): Updating kubeslice webhook to work with new nsm

mohammedzee1000:
Signed-off-by: Mohammed Zeeshan Ahmed <mohammed.zee1000@gmail.com>

### DIFF
--- a/pkg/webhook/pod/webhook.go
+++ b/pkg/webhook/pod/webhook.go
@@ -21,6 +21,7 @@ package pod
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 
 	//	"github.com/kubeslice/worker-operator/controllers"
@@ -39,7 +40,8 @@ const (
 	PodInjectLabelKey                         = "kubeslice.io/pod-type"
 	admissionWebhookSliceNamespaceSelectorKey = controllers.ApplicationNamespaceSelectorLabelKey
 	controlPlaneNamespace                     = "kubeslice-system"
-	nsmInjectAnnotaionKey                     = "ns.networkservicemesh.io"
+	nsmInjectAnnotaionKey1                    = "ns.networkservicemesh.io"
+	nsmInjectAnnotaionKey2                    = "networkservicemesh.io"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/webhook/pod/webhook.go
```diff
@@ -102,7 +103,7 @@ func Mutate(pod *corev1.Pod, sliceName string) *corev1.Pod {
 
 	// Add vl3 annotation to pod template
 	annotations := pod.ObjectMeta.Annotations
-	annotations[nsmInjectAnnotaionKey] = "vl3-service-" + sliceName
+	annotations[nsmInjectAnnotaionKey] = fmt.Sprintf("kernel://vl3-service-%s/nsm0", sliceName)
```
kmjayadeep, line 106: does this work only with kernel forwarder? or will it work with vpp as well?

kmjayadeep:
can you add some UT for checking the annotations, if there isn;t one already?

pkg/webhook/pod/webhook.go
mohammedzee1000: I guess for now atleast our goal is the kernel forwarder. When we support other forwarders, we can look if we need to change this.

mohammedzee1000:
> can you add some UT for checking the annotations, if there isn;t one already?

As the checks have gone green, i am assuming rhere arent any. So I would say we need to add one

pkg/webhook/pod/webhook.go
```diff
@@ -39,7 +40,8 @@ const (
 	PodInjectLabelKey                         = "kubeslice.io/pod-type"
 	admissionWebhookSliceNamespaceSelectorKey = controllers.ApplicationNamespaceSelectorLabelKey
 	controlPlaneNamespace                     = "kubeslice-system"
-	nsmInjectAnnotaionKey                     = "ns.networkservicemesh.io"
+	nsmInjectAnnotaionKey1                    = "ns.networkservicemesh.io"
```
bharath-avesha, line 43: Please fix the spelling of "Annotation".

PR #137, opened 2022-11-16
[WIP] Add support for 1.24+

kmjayadeep:

Adding support for k8s 1.24

Major changes

* Service account token for dashboard is created using tokenrequest api. Due to this [change](https://itnext.io/big-change-in-k8s-1-24-about-serviceaccounts-and-their-secrets-4b909a4af4e0)
* Use the ca cert from `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` to populate the secret in hub since we don't have a secret for serviceaccount anymore

TODO:
- [ ] add some test cases (currently there are no tests for these methods)
- [ ] test in 1.25
- [ ] test in <1.20 (for backward compatibility)

### DIFF
--- a/pkg/hub/hubclient/hubclient.go
+++ b/pkg/hub/hubclient/hubclient.go
@@ -21,10 +21,12 @@ package hub
 import (
 	"context"
 	"errors"
+	"io/ioutil"
 	"os"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
+	authv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -44,9 +46,10 @@ import (
 )
 
 const (
-	GCP   string = "gcp"
-	AWS   string = "aws"
-	AZURE string = "azure"
+	GCP     string = "gcp"
+	AWS     string = "aws"
+	AZURE   string = "azure"
+	CA_FILE        = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
 )
 
 var scheme = runtime.NewScheme()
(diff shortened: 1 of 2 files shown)


pkg/hub/hubclient/hubclient.go
```diff
@@ -143,25 +146,33 @@ func PostClusterInfoToHub(ctx context.Context, spokeclient client.Client, hubCli
...
-		log.Error(err, "Error getting service account")
+
+	if err := spokeclient.Create(ctx, tokenReq); err != nil {
+		log.Error(err, "Error getting service account token")
 		return err
 	}
 
-	secret := &corev1.Secret{}
-	err := spokeclient.Get(ctx, types.NamespacedName{Name: sa.Secrets[0].Name, Namespace: controllers.ControlPlaneNamespace}, secret)
+	ca, err := ioutil.ReadFile(CA_FILE)
```
bharath-avesha, line 164: We are creating a token for the dashboard's service account and using the certs from the worker operator's pod. Can we instead look for a Secret that is bound to the dashboard's service account and read the token and the certs from it? Or create a Secret (i am assuming a token and certs are generated automatically if you create a Secret) and bind it to the service account and read the certs from it?

kmjayadeep: 1. The CA cert is common for the cluster, so it doesn't matter where we read it from.
2. From 1.24, secrets are not automatically created for serviceaccounts, which is why I switched to tokenrequest api. Creating a secret is unnecessary since the only purpose of the secret is to fetch the token once.

pkg/hub/hubclient/hubclient.go
```diff
@@ -143,25 +146,33 @@ func PostClusterInfoToHub(ctx context.Context, spokeclient client.Client, hubCli
 }
 
 func PostDashboardCredsToHub(ctx context.Context, spokeclient, hubClient client.Client) error {
-	sa := &corev1.ServiceAccount{
+	tokenReq := &authv1.TokenRequest{
```
rahulsawra98, line 149: rbac needs to be added for this!

kmjayadeep: nice find! I completely missed this
Maybe it is easier to close this PR and add the serviceaccount with secret in the helm charts itself

rahulsawra98: > nice find! I completely missed this Maybe it is easier to close this PR and add the serviceaccount with secret in the helm charts itself

makes sense!

kmjayadeep:
closing in favor of https://github.com/kubeslice/dev-charts/pull/53

PR #146, opened 2022-12-12
feat(AM-5509): Namespace sameness on a slice 

mridulgain:
- Creating a namespace if that particular namespace is not available in the any cluster which is part of the slice. 
- Prior to this edit, operator reconciler used to throw error, but waited for namespace to be created by user.

### DIFF
--- a/controllers/slice/namespaces.go
+++ b/controllers/slice/namespaces.go
@@ -103,7 +103,7 @@ func (r *SliceReconciler) reconcileAppNamespaces(ctx context.Context, slice *kub
 	// (In case it doesn't exist, first create it & then label the same)
 	// If a namespace is found in the existing list, mark it to indicate that it has been verified
 	// to be valid as it is present in the configured list as well.
-	labeledAppNsList, statusChanged, err := r.labelAppNamespaces(ctx, cfgAppNsList, existingAppNsMap, slice)
+	labeledAppNsList, statusChanged, err := r.createAndLabelAppNamespaces(ctx, cfgAppNsList, existingAppNsMap, slice)
 	if err != nil {
 		return ctrl.Result{}, err, true
 	}
@@ -706,7 +706,7 @@ func buildAppNamespacesList(slice *kubeslicev1beta1.Slice) []string {
 	}
 	return cfgAppNsList
 }
-func (r *SliceReconciler) labelAppNamespaces(ctx context.Context, cfgAppNsList []string, existingAppNsMap map[string]*nsMarker, slice *kubeslicev1beta1.Slice) ([]string, bool, error) {
+func (r *SliceReconciler) createAndLabelAppNamespaces(ctx context.Context, cfgAppNsList []string, existingAppNsMap map[string]*nsMarker, slice *kubeslicev1beta1.Slice) ([]string, bool, error) {
 	labeledAppNsList := []string{}
 	statusChanged := false
 	log := logger.FromContext(ctx).WithValues("type", "appNamespaces")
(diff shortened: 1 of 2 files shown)


controllers/slice/namespaces.go
```diff
@@ -718,9 +719,19 @@ func (r *SliceReconciler) labelAppNamespaces(ctx context.Context, cfgAppNsList [
 		// label does not exists on namespace
 		namespace := &corev1.Namespace{}
 		err := r.Get(ctx, types.NamespacedName{Name: cfgAppNs}, namespace)
+		// If the namespace itself doesn't exist on specified cluster then first create it
 		if err != nil {
```
gourishkb, line 723: Add additional check after err != nil to verify that error is indeed "Failed to find namespaces". 
Like 
`if errors.IsNotFound(err) { 
log.Info("Failed to find Namespaces" , "namespace", cfgAppNs)
.....`

https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsNotFound ref for the function

controllers/slice/namespaces.go
```diff
@@ -718,9 +719,19 @@ func (r *SliceReconciler) labelAppNamespaces(ctx context.Context, cfgAppNsList [
...
-			log.Error(err, "Failed to find namespace", "namespace", cfgAppNs)
-			continue
+			log.Info("Failed to find namespace", "namespace", cfgAppNs)
+			namespace = &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: cfgAppNs,
+				},
+			}
+			err2 := r.Create(ctx, namespace)
```
gourishkb, line 730: Put this in a single if block to avoid "err2". 
``` 
if err := r.Create(ctx, namespace); err!=nil {
```

controllers/slice/namespaces.go
mridulgain: Addressed

controllers/slice/namespaces.go
mridulgain: Addressed

NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2022-12-13T13:35:04-master-245/index.html'

controllers/slice/namespaces.go
```diff
@@ -719,8 +720,23 @@ func (r *SliceReconciler) labelAppNamespaces(ctx context.Context, cfgAppNsList [
...
+			if errors.IsNotFound(err) {
+				log.Info("Namespace is not found. Creating namespace.", "namespace", cfgAppNs)
+				namespace = &corev1.Namespace{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: cfgAppNs,
+					},
+				}
+				if err := r.Create(ctx, namespace); err != nil {
+					log.Error(err, "Failed to create namespace", "namespace", cfgAppNs)
+					continue
```
rahulsawra98, line 733: is there any reason we are continuing in case of error? and not returning error?

mridul-avesha: Suppose there is a scenario where we create a slice CR with app ns array in which we mention an invalid NS name followed by some valid ones. In such case I thought of skipping the invalid one while creating(and labeling) rest of them. That's why I used `continue` instead of `return`. Let me know what you think about this approach?

rahulsawra98 (approved):

NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2022-12-19T06:31:42-master-245/index.html'

NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2022-12-19T07:43:25-master-245/index.html'

PR #159, opened 2022-12-27
Events monitoring package

kmjayadeep:
Part of monitoring framework:

* Created methods and structs to raise events
* As an example, events are raised on nodeip and nodeport update on hub

### DIFF
--- a/pkg/hub/hubclient/hubclient.go
+++ b/pkg/hub/hubclient/hubclient.go
@@ -148,6 +148,7 @@ func (hubClient *HubClientConfig) UpdateNodePortForSliceGwServer(ctx context.Con
 		Message:           fmt.Sprintf("NodePort Updated to: %d", sliceGwNodePort),
 		ReportingInstance: "Controller Reconciler",
 		Object:            sliceGw,
+		Action:            "NodePortUpdated",
 	})
 
 	return err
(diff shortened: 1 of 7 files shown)


pkg/hub/hubclient/hubclient.go
```diff
@@ -129,7 +141,16 @@ func (hubClient *HubClientConfig) UpdateNodePortForSliceGwServer(ctx context.Con
...
 
-	return hubClient.Update(ctx, sliceGw)
+	err = hubClient.Update(ctx, sliceGw)
+	hubClient.eventRecorder.RecordEvent(ctx, &monitoring.Event{
+		EventType:         monitoring.EventTypeNormal,
+		Reason:            monitoring.EventReasonNodePortUpdate,
+		Message:           fmt.Sprintf("NodePort Updated to: %d", sliceGwNodePort),
+		ReportingInstance: "Controller Reconciler",
+		Object:            sliceGw,
+	})
```
gourishkb, line 151: Here we do not report "Action" field but, we do this in the case of NodeIpUpdate above.

gourishkb (approved):
Minor Nitpick, LGTM

rajendra-avesha (approved):

pkg/hub/hubclient/hubclient.go
kmjayadeep: nice find. fixed

PR #165, opened 2023-01-30
Fix istio service discovery reconciler bugs

kmjayadeep:
* Fixed a bug where ServiceImport was showing port number for endpoints as 0.
* Fixed serviceentry was not getting cleaned up after deleting serviceexport with istio ingress/egress gateways


gourishkb (approved):

pkg/hub/hubclient/hubclient.go
```diff
@@ -394,12 +394,15 @@ func getNamespaceInfo(onboardNamespace string, ns []hubv1alpha1.NamespacesConfig
 func getHubServiceDiscoveryEps(serviceexport *kubeslicev1beta1.ServiceExport) []hubv1alpha1.ServiceDiscoveryEndpoint {
 	epList := []hubv1alpha1.ServiceDiscoveryEndpoint{}
 
+	port := serviceexport.Spec.Ports[0].ContainerPort
```
YachikaRalhan, line 397: Is there a possibility that Ports array throws nil pointer error because we are directly accessing the zeroth element

kmjayadeep: ports array should have exactly one element in it ( we don't support multiple ports now). if it is empty, many other things will break so we need to enforce during serviceexport creation it anyway

YachikaRalhan: okay

YachikaRalhan (approved):

PR #188, opened 2023-03-06
feat(): Adding aliases to service export

bharath-avesha:
There is a requirement to identify exported services with arbitrary names instead of slice.local names. This can be achieved by providing an alias field in the service export spec section so that the users can give any name that they desire to the service being exported from a cluster.

Added a new field to hold alias names to the service import and export CRDs.


gourishkb (approved):

gourishkb (approved):

controllers/serviceexport/serviceexport.go
```diff
@@ -229,3 +229,52 @@ func (r *Reconciler) SyncSvcExportStatus(ctx context.Context, serviceexport *kub
...
+			break
+		}
+	}
+
+	if updateNeeded {
+		serviceexport.Status.Aliases = serviceexport.Spec.Aliases
+		serviceexport.Status.LastSync = 0
+		err := r.Status().Update(ctx, serviceexport)
+		if err != nil {
+			log.Error(err, "Failed to update serviceexport ports")
```
Rahul-D78, line 262: Can we have an error message - `Failed to update serviceexport status with Aliases`

api/v1beta1/serviceimport_types.go
```diff
@@ -44,6 +44,9 @@ type ServiceImportSpec struct {
 	DNSName string `json:"dnsName"`
 	// Ports which should be exposed through the service
 	Ports []ServicePort `json:"ports"`
+	// Alias names for the exported service. The service could be addressed by the alias names
+	// in addition to the slice.local name.
+	Aliases []string `json:"aliases,omitempty"`
```
Rahul-D78, line 49: We don't have a function to reconcile Aliases for serviceimport. Where as `ReconcileAliases` function present for serviceexport.

gourishkb:
Do we still need this PR ? @bharath-avesha shall we close it ?

bharath-avesha:
@gourishkb, i have closed it.

PR #193, opened 2023-03-09
feat(AM-7024): Edit node ip

YachikaRalhan:
Support edit node Ip - through this PR we are supporting editing of node IP in controller's cluster CR by updating node Ip's in cluster CR's status
- auto detection of node ip by default on
- remove dependency of node ip in helm values and in operator's env var
- post node ip in cluster CR status in the starting through cluster reconciler
- if there is any change in node, update the node ips in cluster CR through cluster reconciler
- few enhancements in cluster reconciler
- remove slice gateway dependency to validate node ips and update cluster CR
- removed node reconciler dependency, no need of that, as we anyways get node ips in a cluster reconciliation loop


pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -144,14 +143,21 @@ func (r *Reconciler) updateClusterInfo(ctx context.Context, cr *hubv1alpha1.Clus
...
 	log.Info("got clusterinfo", "ci", clusterInfo)
-	if clusterInfo.ClusterProperty.GeoLocation.CloudProvider == GCP || clusterInfo.ClusterProperty.GeoLocation.CloudProvider == AWS || clusterInfo.ClusterProperty.GeoLocation.CloudProvider == AZURE {
-		cr.Spec.ClusterProperty.GeoLocation.CloudRegion = clusterInfo.ClusterProperty.GeoLocation.CloudRegion
-		cr.Spec.ClusterProperty.GeoLocation.CloudProvider = clusterInfo.ClusterProperty.GeoLocation.CloudProvider
-		if err := r.Update(ctx, cr); err != nil {
-			log.Error(err, "Error updating ClusterProperty on hub cluster")
-			return err
+	if clusterInfo.ClusterProperty.GeoLocation.CloudProvider == GCP || clusterInfo.ClusterProperty.GeoLocation.CloudProvider == AWS ||
+		clusterInfo.ClusterProperty.GeoLocation.CloudProvider == AZURE {
+		if clusterInfo.ClusterProperty.GeoLocation.CloudRegion != clusterInfo.ClusterProperty.GeoLocation.CloudRegion ||
```
gourishkb, line 148: is this condition correct , clusterInfo.ClusterProperty.GeoLocation.CloudRegion != clusterInfo.ClusterProperty.GeoLocation.CloudRegion ? maybe a typo ?

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -165,24 +171,35 @@ func (r *Reconciler) updateClusterInfo(ctx context.Context, cr *hubv1alpha1.Clus
...
 		}
+		// TODO raise event to cluster CR that node ip is updated
 	}
-
 	return nil
 }
 
+// total -> external ip list of nodes in the k8s cluster
+// current -> ip list present in nodeIPs of cluster cr
+func validatenodeipcount(total, current []string) bool {
```
gourishkb, line 199: we can probably use a better name for this method like matchNodeIPs() and variable name total can be renamed to something more appropriate

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -165,24 +171,35 @@ func (r *Reconciler) updateClusterInfo(ctx context.Context, cr *hubv1alpha1.Clus
...
+	return reflect.DeepEqual(total, current)
```
rahulsawra98, line 200: reflect.DeepEqual returns false even in case of order mismatch.. so instead we could check the exact elements present in both slices..

rahulsawra98 (changes requested):

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -155,48 +167,116 @@ func (r *Reconciler) updateClusterInfo(ctx context.Context, cr *hubv1alpha1.Clus
 		if err != nil {
 			return err
 		}
-		if clusterInfo.ClusterProperty.GeoLocation.CloudProvider == GCP || clusterInfo.ClusterProperty.GeoLocation.CloudProvider == AWS || clusterInfo.ClusterProperty.GeoLocation.CloudProvider == AZURE {
-			cr.Spec.ClusterProperty.GeoLocation.CloudRegion = clusterInfo.ClusterProperty.GeoLocation.CloudRegion
-			cr.Spec.ClusterProperty.GeoLocation.CloudProvider = clusterInfo.ClusterProperty.GeoLocation.CloudProvider
-			if err := r.Update(ctx, cr); err != nil {
-				log.Error(err, "Error updating ClusterProperty on hub cluster during retry")
-				return err
+		if clusterInfo.ClusterProperty.GeoLocation.CloudProvider == GCP || clusterInfo.ClusterProperty.GeoLocation.CloudProvider == AWS ||
```
bharath-avesha, line 170: please use a var to get the cloudProvider from clusterInfo. The conditional is not easy to read.

cloudProvider := clusterInfo.ClusterProperty.GeoLocation.CloudProvider
if cloudProvider == GCP || cloudProvider == AWS || cloudProvider == AZURE {
}

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -155,48 +167,116 @@ func (r *Reconciler) updateClusterInfo(ctx context.Context, cr *hubv1alpha1.Clus
...
+			clusterInfo.ClusterProperty.GeoLocation.CloudProvider == AZURE {
+			if clusterInfo.ClusterProperty.GeoLocation.CloudRegion != clusterInfo.ClusterProperty.GeoLocation.CloudRegion ||
```
bharath-avesha, line 172: Whats the difference btw the LHS and RHS here?
if clusterInfo.ClusterProperty.GeoLocation.CloudRegion != clusterInfo.ClusterProperty.GeoLocation.CloudRegion
Both are the same.

YachikaRalhan: This is a typo, fixing it

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -38,30 +40,34 @@ type Reconciler struct {
...
-			cr, err = r.getCluster(ctx, req)
-			if err != nil {
-				return reconcile.Result{}, err
-			}
-		}
+	log.Info("got cluster CR from hub", "cluster", cr)
+	cl := cluster.NewCluster(r.MeshClient, cr.Name)
+	res, err, requeue := r.updateClusterCloudProviderInfo(ctx, cr, cl)
+	if err != nil {
+		log.Error(err, "unable to update cloud provider info to cluster")
```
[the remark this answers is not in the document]
kmjayadeep: just a suggestion, what do you think? If too complex we can think about it later

kmjayadeep (approved):
Nice PR :100: 
Thanks for cleaning up and refactoring the old logic. the code looks much better now!

pkg/hub/controllers/cluster/reconciler.go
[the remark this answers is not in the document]
YachikaRalhan: Thankyou for the suggestion JD. We can pick this up separately, as not in scope of this PR.

pkg/cluster/node.go
```diff
@@ -85,19 +83,21 @@ func (n *NodeInfo) populateNodeIpList() error {
 	for i := 0; i < len(nodeList.Items); i++ {
 		nodeIpArr = append(nodeIpArr, nodeList.Items[i].Status.Addresses...)
 	}
+	var cuurentNodeIps []string
 	for i := 0; i < len(nodeIpArr); i++ {
 		if nodeIpArr[i].Type == NodeExternalIP {
-			n.NodeIPList = append(n.NodeIPList, nodeIpArr[i].Address)
+			cuurentNodeIps = append(cuurentNodeIps, nodeIpArr[i].Address)
```
bharath-avesha, line 89: Plz fix the typo in the var name from cuurent to current before merging.

bharath-avesha (approved):

PR #195, opened 2023-03-15
Feature healthstatustimer

gourishkb:
Reconcile cluster CR health status once every 120 seconds and update cluster info immediately when cluster CR is changed.

### DIFF
--- a/pkg/hub/controllers/cluster/reconciler.go
+++ b/pkg/hub/controllers/cluster/reconciler.go
@@ -26,7 +26,7 @@ const (
 	AWS   string = "aws"
 	AZURE string = "azure"
 
-	ReconcileInterval = 10 * time.Second
+	ReconcileInterval = 120 * time.Second
 )
 
 type Reconciler struct {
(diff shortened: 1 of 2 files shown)


rahulsawra98 (approved):

Rahul-D78 (approved):

NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2023-03-15T12:08:31-master-388/index.html'

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -72,6 +73,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	cr.Status.ClusterHealth.LastUpdated = metav1.Now()
 	if err := r.Status().Update(ctx, cr); err != nil {
 		log.Error(err, "unable to update cluster CR")
+		return reconcile.Result{RequeueAfter: ReconcileInterval}, err
```
YachikaRalhan, line 76: This should be return reconcile.Result{Requeue: true}, err, why returning after interval?

PR #196, opened 2023-03-16
fix(): fix health status update conflicts

gourishkb:
Use retry on conflict for updating slice/cluster health status to avoid update errors.

### DIFF
--- a/pkg/hub/controllers/cluster/reconciler.go
+++ b/pkg/hub/controllers/cluster/reconciler.go
@@ -82,10 +82,12 @@ func (r *Reconciler) updateHealthWithRetry(ctx context.Context, cr *hubv1alpha1.
 		if cr.Status.ClusterHealth == nil {
 			cr.Status.ClusterHealth = &hubv1alpha1.ClusterHealth{}
 		}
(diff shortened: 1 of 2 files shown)


richiesebastian (approved):

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -62,23 +61,46 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		}, cr)
+		if err != nil {
+			log.Error(err, "Error while getting cluster CR retrying")
+			return err
+		}
+		if cr.Status.ClusterHealth == nil {
+			cr.Status.ClusterHealth = &hubv1alpha1.ClusterHealth{}
+		}
+		if err := r.updateClusterHealthStatus(ctx, cr); err != nil {
+			log.Error(err, "unable to update cluster health status")
```
YachikaRalhan, line 86: return error

YachikaRalhan: if err := r.updateClusterHealthStatus(ctx, cr); err != nil {
			log.Error(err, "unable to update cluster health status")
this hsould be inside time's condition

PR #203, opened 2023-03-22
fix retry conflicts on health check update

kmjayadeep:
In cluster reconciler, we were not requeueing after updating the CR, which caused conflicts when updating health check. This is not really a big issue, since health check is updated in the next iteration correctly. However a fix is added to avoid unnecessary warnings in the logs

### DIFF
--- a/pkg/hub/controllers/cluster/reconciler.go
+++ b/pkg/hub/controllers/cluster/reconciler.go
@@ -50,7 +50,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if cr.Status.ClusterHealth == nil || time.Since(cr.Status.ClusterHealth.LastUpdated.Time) > 3*time.Minute {
 		log.Info("updating cluster info on controller")
 		if err := r.updateClusterInfo(ctx, cr); err != nil {
+			// Skip the error and continue health check
 			log.Error(err, "unable to update cluster info")
+		} else {
+			return reconcile.Result{Requeue: true}, nil
 		}
 	}
 
@@ -58,6 +61,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if !r.isDashboardCredsUpdated(ctx, cr) {
 		if err := r.updateDashboardCreds(ctx, cr); err != nil {
 			log.Error(err, "unable to update dashboard creds")
+			return reconcile.Result{}, err
+		} else {
+			return reconcile.Result{Requeue: true}, nil
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -50,14 +50,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 		}
 	}
 
 	// Update dashboard creds if it hasn't already
 	if !r.isDashboardCredsUpdated(ctx, cr) {
 		if err := r.updateDashboardCreds(ctx, cr); err != nil {
 			log.Error(err, "unable to update dashboard creds")
+			return reconcile.Result{}, err
+		} else {
+			return reconcile.Result{Requeue: true}, nil
```
gourishkb, line 66: Why not use RequeueAfter: ReconcileInterval here ?

kmjayadeep: ReconcileInterval is used to run health checks periodically. Here we are updating dashboard creds. We would need to reconcile immedietly to continue health check operations

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -50,14 +50,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if cr.Status.ClusterHealth == nil || time.Since(cr.Status.ClusterHealth.LastUpdated.Time) > 3*time.Minute {
 		log.Info("updating cluster info on controller")
 		if err := r.updateClusterInfo(ctx, cr); err != nil {
+			// Skip the error and continue health check
```
Rahul-D78, line 53: Why we are not reconciling if `err != nil` so if update fails due to conflicts or some other reason the cluster info will not populate in the cluster CR and in the UI it will not show any data this could be a bad user experience because without that user could not able to register the cluster. Since it needs nodeIP in the cluster CR spec.

pkg/hub/controllers/cluster/reconciler.go
Rahul-D78: If `updateDashboardCreds` fails why we are not reconciling immedietly ?

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -50,14 +50,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if cr.Status.ClusterHealth == nil || time.Since(cr.Status.ClusterHealth.LastUpdated.Time) > 3*time.Minute {
```
Rahul-D78, line 50: So if in the first run the nodeIP update fails in that case, we have to wait for next 3 min to populate back to the cluster CR. can we reduce the time ?

pkg/hub/controllers/cluster/reconciler.go
kmjayadeep: `return reconcile.Result{}, err` requeues immedietly if error is passed. no need to specify requeue: true

gourishkb (approved):
LGTM

kmjayadeep:
don't merge yet. testing in progress

PR #207, opened 2023-03-31
fix(): fixed slicegw health status for single cluster

gourishkb:
For a single cluster, sliceGW pods are missing.
We need to handle component health status for two cases 1. single cluster 2. multiple cluster

case 1: we check if sliceGW deployment is present, expected no slicegw deployments hence, mark healthystate

case 2: we check if sliceGW deployment is present, expected 1 slicegw deployment, hence, list all pods and check if all are in Running state else mark as unhealthyState

### DIFF
--- a/pkg/hub/controllers/slice_controller.go
+++ b/pkg/hub/controllers/slice_controller.go
@@ -437,6 +437,11 @@ func (r *SliceReconciler) fetchSliceGatewayHealth(ctx context.Context, c *compon
 				cs.ComponentHealthStatus = spokev1alpha1.ComponentHealthStatusError
 				return cs, nil
 			}
+			if len(pods) != len(sliceGwDeployments.Items) {
+				log.Error(fmt.Errorf("number of pods do not match slicegw deployments running"), "unhealthy", "pod", c.name)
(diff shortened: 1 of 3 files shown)


kmjayadeep (approved):
tbh, this feels like a hacky fix for slicegateway. I think we should revisit this later

NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2023-04-01T14:04:02-release-guwahati-394/index.html'

NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2023-04-01T16:48:11-release-guwahati-394/index.html'

gourishkb:
https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2023-04-01T16:48:11-release-guwahati-394/index.html tests has AfterAll suite and other tests failing are seen without this PR change as well as observed here https://github.com/kubeslice/kubeslice-e2e-automation/actions/runs/4589005770/jobs/8103599087. So merging this PR.

PR #209, opened 2023-04-01
fix(): nodeip update while empty string

Rahul-D78:
### This PR contains following fixes:
* If cluster.spec.nodeIPs contains empty string []string{""} then worker operator is not able to update the nodeIPs.
* removed the check where we are checking if the `ClusterHealth == nil` this condition will be always true after the first reconcilation due to which in case there is a nodeIP change detected reconciler will never call the `updateClusterInfo` function.

### DIFF
--- a/pkg/hub/controllers/cluster/reconciler.go
+++ b/pkg/hub/controllers/cluster/reconciler.go
@@ -166,7 +166,7 @@ func (r *Reconciler) updateClusterInfo(ctx context.Context, cr *hubv1alpha1.Clus
 
 	// Populate NodeIPs if not already updated
 	// Only needed to do the initial update. Later updates will be done by node reconciler
-	if isEmptyString(cr.Spec.NodeIPs) || cr.Spec.NodeIPs == nil || len(cr.Spec.NodeIPs) == 0 {
+	if isValidNodeIpList(cr.Spec.NodeIPs) || cr.Spec.NodeIPs == nil || len(cr.Spec.NodeIPs) == 0 {
 		nodeIPs, err := cluster.GetNodeIP(r.MeshClient)
 		if err != nil {
 			log.Error(err, "Error Getting nodeIP")
@@ -270,7 +270,7 @@ func (r *Reconciler) getCluster(ctx context.Context, req reconcile.Request) (*hu
 	return hubCluster, nil
 }
 
-func isEmptyString(nodeIPs []string) bool {
+func isValidNodeIpList(nodeIPs []string) bool {
 	for _, nodeIP := range nodeIPs {
 		if nodeIP == "" {
 			return true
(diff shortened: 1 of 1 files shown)


gourishkb (approved):
LGTM

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -272,6 +270,15 @@ func (r *Reconciler) getCluster(ctx context.Context, req reconcile.Request) (*hu
 	return hubCluster, nil
 }
 
+func isEmptyString(nodeIPs []string) bool {
+	for _, nodeIP := range nodeIPs {
+		if nodeIP == "" {
+			return true
+		}
```
kmjayadeep, line 277: So if the array is ["10.1.2.2",""]. then this method returns true. is it really what is intended?

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -272,6 +270,15 @@ func (r *Reconciler) getCluster(ctx context.Context, req reconcile.Request) (*hu
 	return hubCluster, nil
 }
 
+func isEmptyString(nodeIPs []string) bool {
```
kmjayadeep, line 273: The method name is confusing. How about we invert the logic and call it `isValidNodeIpList` or `isNonEmptyArray`?

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -46,12 +46,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 	log.Info("got cluster CR from hub", "cluster", cr)
 
-	// Post NodeIP and GeoLocation info only on first run or if the reconciler wasn't run for a while
-	if cr.Status.ClusterHealth == nil || time.Since(cr.Status.ClusterHealth.LastUpdated.Time) > 3*time.Minute {
-		log.Info("updating cluster info on controller")
-		if err := r.updateClusterInfo(ctx, cr); err != nil {
-			log.Error(err, "unable to update cluster info")
-		}
+	// Post NodeIP and GeoLocation info
+	log.Info("updating cluster info on controller")
```
kmjayadeep, line 50: The above block and the node controller were already removed by the PR  from @YachikaRalhan in https://github.com/kubeslice/worker-operator/pull/193. Maybe we don't need this PR if those changes are taken from master. Please take a look

kmjayadeep (changes requested):
The above block and the node controller were already removed by the PR from @YachikaRalhan in https://github.com/kubeslice/worker-operator/pull/193. Maybe we don't need this PR if those changes are taken from master. Please take a look

pkg/hub/controllers/cluster/reconciler.go
Rahul-D78: Yes this is intended. so we are going to range through all the nodeIPs and if we found empty string we will return true.

PR #212, opened 2023-04-03
raise events for edit node ip

YachikaRalhan:
- Implemented new monitoring framework changes, where events can be added in the component itself, no external dependency
- Raising two events when updating node ip in cluster cr for auto detection
   - ClusterNodeIpAutoDetectionFailed
   - ClusterNodeIpAutoDetected
- Additionally raising correct error when there is an error updating nsm cni subnet in cluster cr to reconcile


pkg/cluster/cluster.go
```diff
@@ -112,17 +113,17 @@ func (c *Cluster) GetNsmExcludedPrefix(ctx context.Context, configmap, namespace
 	}
 	if len(nsmconfig.Data) == 0 {
 		log.Info("prefix data not present")
```
rahulsawra98, line 115: @YachikaRalhan , can we please also change this log to log.Error ?

pkg/cluster/cluster.go
```diff
@@ -112,17 +113,17 @@ func (c *Cluster) GetNsmExcludedPrefix(ctx context.Context, configmap, namespace
...
-		return nil, err
+		return nil, errors.New("prefix data not present in nsm configmap")
 	}
 	_, ok := nsmconfig.Data["excluded_prefixes_output.yaml"]
 	if !ok {
 		log.Info("cni subnet info not present")
-		return nil, err
+		return nil, errors.New("cni subnet info not present in nsm configmap")
```
rahulsawra98, line 121: same change as above!

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -286,10 +286,25 @@ func (r *Reconciler) updateNodeIps(ctx context.Context, cr *hubv1alpha1.Cluster)
 	})
 	if err != nil {
 		log.Error(err, "Error updating to node ip's on hub cluster")
+		err := r.EventRecorder.RecordEvent(ctx, &events.Event{
```
rahulsawra98, line 289: why is this event needed? 
Failing to update is a normal error in controller's pattern,
This can give a wrong interpretation to user , because in next reconcile loop it can succeed.
And this is an "internal" event - details should not be exposed to user!

rahulsawra98 (changes requested):

rahulsawra98:
@YachikaRalhan I strongly believe the events exposed as a part of this PR are internal (developer) events which should be "debug" logs and not events.
A "user" of kubeslice does not know what "updating/posting" nodeIP to cluster CR means.
Please reconsider this!

pkg/hub/controllers/cluster/reconciler.go
YachikaRalhan: We have discussed this case, so we want to raise an event if there is a failure and in the next reconcile loop if the update is successful we raise a new event that now update has happened. So the latest event will be successful.

Rahul-D78 (approved):

PR #213, opened 2023-04-03
fix(): slicegw health status for single cluster

Rahul-D78:
For a single cluster, sliceGW pods are missing.
We need to handle component health status for two cases 1. single cluster 2. multiple cluster

case 1: we check if sliceGW deployment is present, expected no slicegw deployments hence, mark healthystate

case 2: we check if sliceGW deployment is present, expected 1 slicegw deployment, hence, list all pods and check if all are in Running state else mark as unhealthyState

### DIFF
--- a/pkg/hub/controllers/cluster/reconciler.go
+++ b/pkg/hub/controllers/cluster/reconciler.go
@@ -276,7 +276,7 @@ func (r *Reconciler) updateNodeIps(ctx context.Context, cr *hubv1alpha1.Cluster)
 		// the current nodeIpList (nodeIpList contains list of externalIPs or internalIPs)
 		// if the nodeIP is no longer available, we update the cluster CR on controller cluster
 		if cr.Status.NodeIPs == nil || len(cr.Status.NodeIPs) == 0 ||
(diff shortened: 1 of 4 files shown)


gourishkb (approved):

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -276,7 +276,7 @@ func (r *Reconciler) updateNodeIps(ctx context.Context, cr *hubv1alpha1.Cluster)
 		// the current nodeIpList (nodeIpList contains list of externalIPs or internalIPs)
 		// if the nodeIP is no longer available, we update the cluster CR on controller cluster
 		if cr.Status.NodeIPs == nil || len(cr.Status.NodeIPs) == 0 ||
-			!validatenodeips(nodeIPs, cr.Status.NodeIPs) {
+			!validatenodeips(nodeIPs, cr.Status.NodeIPs) || !isValidNodeIpList(nodeIPs) {
 			log.Info("Mismatch in node IP", "IP in use", cr.Status.NodeIPs, "IP to be used", nodeIPs)
```
YachikaRalhan, line 280: why is this needed? validatenodeips function is already returning false if the elements of array are not same.

PR #221, opened 2023-04-12
fix(): edit nodeIP fixes, add aliases to service export

Rahul-D78:
This PR includes edit nodeIP fixes and aliases support:

* Filtering slicegw pods using slicegw pod label.

* There is a requirement to identify exported services with arbitrary names
* There is a requirement to identify exported services with arbitrary names instead of slice.local names. This can be achieved by providing an alias field in the service export spec section so that the users can give any name that they desire to the service being exported from a cluster.
* Added a new field to hold alias names to the service import and export CRDs.

Signed-off-by: Bharath Horatti <bharath@aveshasystems.com>
Signed-off-by: YachikaRalhan <yachika.ralhan@aveshasystems.com>
Signed-off-by: Rahul-D78 <rahul.kumar@aveshasystems.com>


rahulsawra98 (approved):

YachikaRalhan (approved):

kmjayadeep:
Please add a description
Where are the aliases used? I only see the changes to add the field to CRD

PR #225, opened 2023-04-21
Fix alias update in ServiceExport

bharath-avesha:
Bug fix for https://avesha.atlassian.net/browse/AM-7487

Serviceexport reconciliation was failing when new items to the list of aliases were added.


controllers/serviceexport/serviceexport.go
```diff
@@ -242,13 +242,18 @@ func (r *Reconciler) ReconcileAliases(
...
 
-	// Status does not contain alias info. Could be the first iteration of the reconcile loop
-	if len(serviceexport.Status.Aliases) == 0 {
-		updateNeeded = true
-	}
+	// There is no do..while in Go, using the 'for' loop to achieve do..while here. It helps to keep the code
+	// clean in cases like these where the control flow needs to jump (like the goto statement in C) to an 
+	// instruction deeper down the function when a condition is satisfied and there is no need to test any 
+	// other conditions.
+	for doOnce := true; doOnce; doOnce = false {
```
rahulsawra98, line 249: If I understood correctly, you re running the for loop only once.
1. Why do you need for loop exactly?
2. Why do you need break statement?

rahulsawra98, line 249: If I understand correctly , this for loop would be executed only once.
why do you need for loop then?

gourishkb: if I understand this correctly, we are doing two checks (length mismatch of aliases and if alias is mismatched) and only then deciding to update. @bharath-avesha why not use a function call to check for both conditions instead of this or use a goto statement as you mentioned in the comment.

bharath-avesha: @gourishkb right now there are two conditionals, but it started off with a couple more. I guess i can merge the two into a single func call. We should be doing a do..while(0) equivalent if and when we add more conditionals. I prefer using do..while(0) with break statements instead of goto because the flow of the function is interrupted in the latter and does not make a good code reading experience especially if there are multiple gotos.

PR #229, opened 2023-04-21
fix(): update cluster health after ReconcileInterval 

Rahul-D78:


priyank-upadhyay (approved):
Looks good

richiesebastian (approved):

rajendra-avesha (approved):
@Rahul-D78 are we updating the cluster health continously without periodic update?

Rahul-D78:
> @Rahul-D78 are we updating the cluster health continously without periodic update?

@rajendra-avesha we missed to pull this fix from `release-guhwati` branch.
Without this fix there is a race condition.

PR #238, opened 2023-04-26
feat(): cluster auto deregister

Rahul-D78:
Through this PR we are supporting cluster auto deregistration by
* Adding worker finalizer to cluster CR
* Cluster reconciler detects finalizer and deletion timestamp
* Create a job to uninstall helm chart
* Notify controller about registration status
* Remove worker finalizer after clenup job succeeded


config/events/worker-events.yaml
```diff
@@ -28,4 +28,10 @@ events:
     action: None
     type: Warning
     reportingController: worker
-    message: Auto-detection of cluster node IP addresses failed
\ No newline at end of file
+    message: Auto-detection of cluster node IP addresses failed
+  - name: DeregitrationJobFailed
```
rajendra-avesha, line 32: Please check for Deregitration spelling Deregistration

events/events_generated.go
```diff
@@ -62,6 +62,14 @@ var EventsMap = map[events.EventName]*events.EventSchema{
 		ReportingController: "worker",
 		Message:             "Auto-detection of cluster node IP addresses failed",
 	},
+	"DeregitrationJobFailed": {
```
rajendra-avesha, line 65: DeregitrationJobFailed change to DeregistrationJobFailed

events/events_generated.go
```diff
@@ -70,4 +78,5 @@ var (
 	EventClusterHealthy                   events.EventName = "ClusterHealthy"
 	EventClusterNodeIpAutoDetected        events.EventName = "ClusterNodeIpAutoDetected"
 	EventClusterNodeIpAutoDetectionFailed events.EventName = "ClusterNodeIpAutoDetectionFailed"
+	EventDeregitrationJobFailed           events.EventName = "DeregitrationJobFailed"
```
rajendra-avesha, line 81: DeregitrationJobFailed to DeregistrationJobFailed

pkg/hub/controllers/cluster/deregister.go
```diff
@@ -0,0 +1,330 @@
...
+		cluster.Status.RegistrationStatus = status
+		return r.Status().Update(ctx, cluster)
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *Reconciler) setIsDeregisterInProgress(ctx context.Context, cluster *hubv1alpha1.Cluster, isDeregisterInProgress bool) error {
```
rajendra-avesha, line 170: setIsDeregisterInProgress method seems to be un used Do you really need this?

pkg/hub/controllers/cluster/deregister.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	return clusterRole
+}
+
+func getConfigmapScriptPath(file string) string {
+	dir := os.Getenv("SCRIPT_PATH")
+	if dir != "" {
+		return path.Join(dir, file+".sh")
+	}
+	return path.Join("../../../../scripts", file+".sh")
```
rajendra-avesha, line 242: Can you please have this "../../../../scripts" path as a constant

pkg/hub/controllers/cluster/deregister.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func getConfigmapData() (string, error) {
+	fileName := "cleanup"
```
rajendra-avesha, line 246: Can we club both getConfigmapData and getConfigmapScriptPath to single method?

pkg/hub/controllers/cluster/deregister.go
```diff
@@ -0,0 +1,330 @@
...
+			TTLSecondsAfterFinished: &ttlSecondsAfterFinished,
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: deregisterJobName,
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName: serviceAccountName,
+					Containers: []corev1.Container{{
+						Name:  serviceAccountName,
+						Image: "aveshadev/worker-installer:latest",
```
rajendra-avesha, line 288: Is this image going to constant both in dev and prod? OSS and ENT?

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -282,6 +309,7 @@ func (r *Reconciler) updateCNISubnetConfig(ctx context.Context, cr *hubv1alpha1.
 		}
 		if !reflect.DeepEqual(cr.Status.CniSubnet, cniSubnet) {
 			cr.Status.CniSubnet = cniSubnet
+			cr.Status.RegistrationStatus = hubv1alpha1.RegistrationStatusRegistered
```
rajendra-avesha, line 312: Are we still checking for CniSubnet to set RegistrationStatusRegistered
Can't we check for Cluster Health check to set RegistrationStatusRegistered

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -458,3 +486,47 @@ func (r *Reconciler) InjectClient(c client.Client) error {
...
+				log.Error(err, "unable to deregister the worker operator")
+				// increment count for retryAttempts
+				retryAttempts++
+				// resetting isDeregisterInProgress to false
+				statusUpdateErr := r.updateRegistrationStatusAndDeregisterInProgress(ctx, cluster, hubv1alpha1.RegistrationStatusDeregisterFailed, false)
+				if statusUpdateErr != nil {
+					// log.Error(statusUpdateErr, "unable to update registration status")
+					recordEventErr := r.EventRecorder.RecordEvent(ctx, &events.Event{
+						Object:            cluster,
+						Name:              ossEvents.EventDeregitrationJobFailed,
```
rajendra-avesha, line 518: globally please correct EventDeregitrationJobFailed to EventDeregistrationJobFailed

pkg/hub/controllers/cluster/reconciler_test.go
```diff
@@ -0,0 +1,284 @@
...
+		mock.IsType(ctx),
+		mock.IsType(clusterKey),
+		mock.IsType(&hubv1alpha1.Cluster{}),
+	).Return(nil)
+	client.StatusMock.On("Update",
+		mock.IsType(ctx),
+		mock.IsType(&hubv1alpha1.Cluster{}),
+		mock.IsType([]k8sclient.UpdateOption(nil)),
+	).Return(errors.New("error updating status of deregistration on the controller"))
+	client.On("Create",
```
rajendra-avesha, line 274: Haven't we written any ITs for Cluster Deregister?
Please add the same

pkg/hub/controllers/cluster/deregister.go
```diff
@@ -0,0 +1,330 @@
...
+		cluster.Status.IsDeregisterInProgress = isDeregisterInProgress
+		return r.Status().Update(ctx, cluster)
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *Reconciler) updateClusterDeregisterStatus(status hubv1alpha1.RegistrationStatus, ctx context.Context, cluster *hubv1alpha1.Cluster) error {
```
rajendra-avesha, line 152: Can we club both updateRegistrationStatusAndDeregisterInProgress and updateClusterDeregisterStatus methods
As updateRegistrationStatusAndDeregisterInProgress is sub status of the updateClusterDeregisterStatus

rajendra-avesha (changes requested):

pkg/hub/controllers/cluster/deregister.go
Rahul-D78: It's easier for Unit test cases. that is why we have isolated the functions.

pkg/hub/controllers/cluster/reconciler_test.go
Rahul-D78: We have written UT's we are getting a good coverage (90%). and as disscussed over call we can not write IT's with the current bash script approach.
We are testing the whole flow using e2e and e2e with UI too. Apart from this we have fixed all the current ITs also.

pkg/hub/controllers/cluster/reconciler.go
Rahul-D78: Currently the approach to validate the cluster is registered or not is through CNI subnet available. UI also requires CNI subnet to set status as registered. The slice creation also requires this field to be set. health check current only perform component health checks and doesn't provide overall healthcheck.

pkg/hub/controllers/cluster/deregister.go
Rahul-D78: Yes it is an opensource image.

pkg/hub/controllers/cluster/deregister.go
Rahul-D78: We tried clubbing the functions together tests are failing because of the same function name. so currently we have segregated functions.

pkg/hub/controllers/cluster/reconciler_test.go
rajendra-avesha: @Rahul-D78 We can have IT atleast till JOB creation

pkg/hub/controllers/cluster/reconciler.go
rajendra-avesha: @kmjayadeep @pnavali @bharath-avesha Can you please confirm on this?

pkg/hub/controllers/cluster/deregister.go
rajendra-avesha: Is this pointing to dev or Prod repo?

pkg/hub/controllers/cluster/deregister.go
rajendra-avesha: I don't think this is right approach I think we can write UTs for a method of many lines

pkg/hub/controllers/cluster/deregister.go
rajendra-avesha: I don't think this is right approach I think we can write UTs for a method of many lines

pkg/hub/controllers/cluster/reconciler_test.go
YachikaRalhan: We tried writing the IT until job creation but that is not happening because the envtest job reconciler is not adding unique selector while creation of job which is mandatory for a job to be created. If we manually add the selector in job spec the test case is passing but manually adding the selector in job spec is not a recomended because it means we have to select a unique label on our own for the job and if we fail in doing such thing it may harm other jobs too.
It is mentioned here in the k8s docs to not mention selector spec - https://kubernetes.io/docs/concepts/workloads/controllers/job/#specifying-your-own-pod-selector

pkg/hub/controllers/cluster/deregister.go
Rahul-D78: Sorry I didn't get your point. This is an opensource docker image publicly accessible. This only contains an alpine image, with helm, kubectl and some general utilities jq and yq.

pkg/hub/controllers/cluster/deregister.go
Rahul-D78: We can do it. but is it necessary ?

pkg/hub/controllers/cluster/reconciler_test.go
YachikaRalhan: Added a integration test case

pkg/hub/controllers/cluster/reconciler.go
YachikaRalhan: We will move this move of code once health check is implemented for cni subnet and node ips (going as tech debt) @rajendra-avesha

scripts/cleanup.sh
```diff
@@ -0,0 +1,84 @@
...
+SECRET=kubeslice-hub
+TOKEN=$(kubectl get secret ${SECRET} -o json | jq -Mr '.data.token' | base64 -d)
+kubectl get secret ${SECRET} -o json | jq -Mr '.data["ca.crt"]' | base64 -d > /ca.crt
+# get the worker release name
+workerRelease=$(helm list --output json -n kubeslice-system | jq -r '.[] | select(.chart | startswith("kubeslice-worker-")).name')
+echo "🏃 Running helm uninstall for the release $workerRelease" 
+if helm uninstall $workerRelease --namespace kubeslice-system
+then
+    # delete crds
+    deleteKubeSliceCRDs
```
bharath-avesha, line 35: Can we please move all these commands into a func and reuse it in the else clause below?

if helm_uninstall_succeeded
then
    run_post_uninstall_cleanup
else
   if helm_uninstall_no_hook_succeeded
   then
       run_forced_post_uninstall_cleanup


run_forced_post_uninstall_cleanup(){
    # Do forced ops here
    run_post_uninstall_cleanup
}

pkg/hub/controllers/cluster/deregister.go
```diff
@@ -0,0 +1,302 @@
...
+			log.Info("cluster deregister job already exists", "job", clusterDeregisterConfigMap)
+		} else {
+			log.Error(err, "unable to create job for cluster deregister")
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *Reconciler) updateRegistrationStatusAndDeregisterInProgress(ctx context.Context, cluster *hubv1alpha1.Cluster, status hubv1alpha1.RegistrationStatus, isDeregisterInProgress *bool) error {
```
bharath-avesha, line 138: Not sure if i understand the need for the "isDeregisterInProgress *bool" param. A couple of questions:

Why does it have to be a pointer? 
From the code, it is only set to false in the handleDeletion func in the other file. Dont see the flag being set to true. When is it set to true?
Why cant the hubv1alpha1.RegistrationStatusDeregisterInProgress be used to check the status instead of having this separate flag?

pkg/hub/controllers/cluster/deregister.go
```diff
@@ -0,0 +1,302 @@
...
+	if err := r.MeshClient.Create(ctx, constructClusterRoleBinding(), &client.CreateOptions{}); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			log.Info("cluster rolebinding already exists", "clusterrolebinding", clusterRoleBindingName)
+		} else {
+			log.Error(err, "unable to create cluster rolebinding")
+			return err
+		}
+	}
+	// get configmap data
+	data, err := getConfigmapData()
```
bharath-avesha, line 96: getConfigmapData() is confusing. It could mean there is some configmap that needs to be read. To be clear, i would suggest changing the name to getCleanupScript(), and write a comment: 
// get the cleanup script in a long string and construct a configmap from it

pkg/hub/controllers/cluster/deregister.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		log.Error(err, "unable to fetch configmap data")
+		return err
+	}
+	// Set up a configuration map which contains the script for the job.
+	if err := r.MeshClient.Create(ctx, constructConfigMap(data), &client.CreateOptions{}); err != nil {
```
bharath-avesha, line 102: I think it would be cleaner to delete the configmap if it is already present.

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -458,3 +488,48 @@ func (r *Reconciler) InjectClient(c client.Client) error {
...
+					ReportingInstance: "cluster_reconciler",
+				})
+				if recordEventErr != nil {
+					log.Error(recordEventErr, "unable to record event for cluster deregistration")
+				}
+				return true, reconcile.Result{}, err
+			}
+		}
+		// Stop reconciliation as the item is being deleted
+		return true, reconcile.Result{}, nil
```
bharath-avesha, line 532: Should there be a timeout to reconcile.Result{} since we are requeuing? It will immediately trigger the reconciler if you are requeing w/o a timer. It would be wasteful unnecessary func calls in this case.

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -458,3 +488,48 @@ func (r *Reconciler) InjectClient(c client.Client) error {
...
+		// registering our finalizer.
+		if !controllerutil.ContainsFinalizer(cluster, clusterDeregisterFinalizer) {
+			controllerutil.AddFinalizer(cluster, clusterDeregisterFinalizer)
+			if err := r.Update(ctx, cluster); err != nil {
+				return true, reconcile.Result{}, err
+			}
+		}
+	} else {
+		// The object is being deleted
+		if controllerutil.ContainsFinalizer(cluster, clusterDeregisterFinalizer) && !cluster.Status.IsDeregisterInProgress && retryAttempts <= 3 {
```
bharath-avesha, line 507: Cant we rely on the deregister status enum defined by hubv1alpha1 instead of the IsDeregisterInProgress flag?

bharath-avesha, line 507: Lets have a global const to define the retry attempts.

MAX_CLUSTER_DEREGISTRATION_ATTEMPTS = 4

pkg/hub/controllers/cluster/deregister.go
```diff
@@ -0,0 +1,302 @@
...
+	cleanupContainer           = "cluster-cleanup"
+	clusterRoleName            = "cluster-deregister-role"
+	clusterRoleBindingName     = "cluster-deregister-rb"
+	clusterDeregisterConfigMap = "cluster-deregister-cm"
+	deregisterJobName          = "cluster-deregisteration-job"
+	operatorClusterRoleName    = "kubeslice-manager-role"
+)
+
+const (
+	scriptPath = "../../../../scripts"
```
bharath-avesha, line 49: Lets use the full path instead of relative path.

pkg/hub/controllers/cluster/deregister.go
YachikaRalhan: Combined the function

pkg/hub/controllers/cluster/deregister.go
YachikaRalhan: It will impact controller side if we remove isDeregisterInProgress bool. @priyank-upadhyay could please give a little explanation why it is needed.

YachikaRalhan: And pointer because we don't want to set it true, controller will do it and to use same function to update RegistrationStatus and IsDeregisterInProgress we wrote it like that @bharath-avesha

pkg/hub/controllers/cluster/reconciler.go
priyank-upadhyay: The `IsDeregisterInProgress` flag is set to `true` by controller to disable redundant event triggers and requeues of cluster reconciler in the controller.
The worker operator is simply utilizing the flag to retry 3-4 times before failing. Post which the controller will take care of removing the finalizer (after 10 mins)

We could use the enums, but the problem is to identify (at controller sied) if it is the first attempt or a retry which would not be possible if relying on the enums.

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -458,3 +489,50 @@ func (r *Reconciler) InjectClient(c client.Client) error {
...
+		// The object is being deleted
+		if controllerutil.ContainsFinalizer(cluster, clusterDeregisterFinalizer) && !cluster.Status.IsDeregisterInProgress && retryAttempts <= MAX_CLUSTER_DEREGISTRATION_ATTEMPTS {
+			// our finalizer is present, so lets handle any external dependency
+			if err := r.createDeregisterJob(ctx, cluster); err != nil {
+				// unable to deregister the worker operator, return with an error and raise event
+				log.Error(err, "unable to deregister the worker operator")
+				// increment count for retryAttempts
+				retryAttempts++
+				// resetting isDeregisterInProgress to false
+				isDeregisterInProgress := false
```
bharath-avesha, line 516: @priyank-upadhyay @YachikaRalhan 
From what i understood from Priyank's explanation is that the IsDeregisterInProgress is used by the controller to maintain some internal state, and it is also being used by the operator to retry the cleanup. But the operator is relying on the retryAttempts var here to check if another attempt needs to be made. The control will reach this point in the code only if the cluster.Status.IsDeregisterInProgress is false (see line 508) and we are again setting it to false and updating the status. So i am not sure why it needs to be set to false by the operator, since its a no-op as the cluster.Status.IsDeregisterInProgress would anyways be false. IMO the controller and operator should not be modifying the same elements in the Status section. The operator should set the hubv1alpha1.RegistrationStatus and the controller should read it and decide what to do next. But we should not have both operator and controller writing to hubv1alpha1.RegistrationStatus. Should be readonly for one and readWrite for the other.

pkg/hub/controllers/cluster/reconciler.go
```diff
@@ -458,3 +489,50 @@ func (r *Reconciler) InjectClient(c client.Client) error {
...
+		// registering our finalizer.
+		if !controllerutil.ContainsFinalizer(cluster, clusterDeregisterFinalizer) {
+			controllerutil.AddFinalizer(cluster, clusterDeregisterFinalizer)
+			if err := r.Update(ctx, cluster); err != nil {
+				return true, reconcile.Result{}, err
+			}
+		}
+	} else {
+		// The object is being deleted
+		if controllerutil.ContainsFinalizer(cluster, clusterDeregisterFinalizer) && !cluster.Status.IsDeregisterInProgress && retryAttempts <= MAX_CLUSTER_DEREGISTRATION_ATTEMPTS {
```
bharath-avesha, line 508: MAX_CLUSTER_DEREGISTRATION_ATTEMPTS is set to 3. But here its '<=' which means it will attempt 4 times. Should this run 3 or 4 attempts? If 3 then it should be '<', not '<='.

bharath-avesha (approved):

YachikaRalhan (approved):

richiesebastian:
merging for @Rahul-D78

Rahul-D78:
@richiesebastian can you please merge this PR since we are not able to merge because of this error.
```
The base branch requires all commits to be signed
```

PR #247, opened 2023-05-10
Fix labels for metrics

kmjayadeep:
* namespace was renamed to slice_namespace to avoid confusion
* use correct slice name for slice health check metrics

### DIFF
--- a/controllers/slice/reconciler.go
+++ b/controllers/slice/reconciler.go
@@ -386,7 +386,7 @@ func isAppPodStatusChanged(current []kubeslicev1beta1.AppPod, old []kubeslicev1b
 // Setup SliceReconciler
 // Initializes metrics and sets up with manager
 func (r *SliceReconciler) Setup(mgr ctrl.Manager, mf metrics.MetricsFactory) error {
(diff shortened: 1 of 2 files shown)


controllers/slice/reconciler.go
```diff
@@ -386,7 +386,7 @@ func isAppPodStatusChanged(current []kubeslicev1beta1.AppPod, old []kubeslicev1b
 // Setup SliceReconciler
 // Initializes metrics and sets up with manager
 func (r *SliceReconciler) Setup(mgr ctrl.Manager, mf metrics.MetricsFactory) error {
-	gaugeAppPods := mf.NewGauge("app_pods", "App pods in slice", []string{"slice", "namespace"})
+	gaugeAppPods := mf.NewGauge("app_pods", "App pods in slice", []string{"slice", "slice_namespace"})
```
YachikaRalhan, line 389: What confusion is here btw? @kmjayadeep

YachikaRalhan (approved):

controllers/slice/reconciler.go
kmjayadeep: prometheus adds a few labels by default, which includes `pod` name as the operator pod name (source of the metric). So it looks like this namespace is the namespace where the operator runs.

PR #248, opened 2023-05-10
Fix metric labels

kmjayadeep:
* namespace was renamed to slice_namespace to avoid confusion
* use correct slice name for slice health check metrics


gourishkb:
This PR is a duplicate of https://github.com/kubeslice/worker-operator/pull/247. Closing the PR

PR #260, opened 2023-06-09
fix(): worker events reference fixed

gourishkb:
1. removed older event causing errors
2. added reference to events


Rahul-D78 (approved):

gourishkb:
NA

PR #265, opened 2023-06-21
fix(): Changed istio serviceentry resolution to static

bharath-avesha:
Changed the istio serviceentry resolution to static from dns when setting up the istio egress gw. There is a problem with dns resolution when the istio-proxy is run in the router mode wherein the resolution of hostnames fails intermittently and in some cases it remains in the failed state for a very long time. Interestingly, this issue is not seen with istio-proxy running in sidecar mode.
We use static resolution for istio ingress gw, following the same practice for egress gw as well.

### DIFF
--- a/.github/workflows/delete-old-workflow-manual.yaml
+++ b/.github/workflows/delete-old-workflow-manual.yaml
@@ -1,0 +1,59 @@
+name: Delete old workflow runs manual
+on:
+  workflow_dispatch:
+    inputs:
+      days:
+        description: 'Number of days.'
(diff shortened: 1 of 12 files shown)


NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2023-06-29T06:00:49-release-mumbai-414/index.html'

rajendra-avesha (approved):

rajendra-avesha (approved):

PR #271, opened 2023-07-11
fix(): fix gateway rebalancing

gourishkb:
Use nodeCount by making kubernetes API call,
find count of all the nodes with kubeslice.io/node-type=gateway. 
Instead of relying on the node reconciler to update the NodeExternalIPList (and using that as the nodeCount)


rajendra-avesha (approved):

Rahul-D78 (approved):

rahulsawra98 (approved):
Looks good!

PR #275, opened 2023-07-24
fix(): fix fsm bugs and refactor

rahulsawra98:


pkg/hub/controllers/workerslicegwrecycler/utils.go
```diff
@@ -241,10 +252,10 @@ func (r *Reconciler) update_routing_table(e *fsm.Event) error {
 			return errors.New("route not yet present")
 		}
 		return nil
-	}, retry.Attempts(100))
+	}, retry.Attempts(200))
 
 	if err != nil {
-		log.Error(err, "Err():", err.Error())
+		log.Error(err, "Err(): moving FSM to error state:", err.Error())
```
YachikaRalhan, line 258: This should be - `log.Error(err, "Err(): moving FSM to error state:") `in the end it expects key/values

pkg/hub/controllers/workerslicegwrecycler/utils.go
```diff
@@ -260,7 +271,7 @@ func (r *Reconciler) update_routing_table(e *fsm.Event) error {
 		err = retry.Do(func() error {
 			workerslicegwrecycler.Status.Client.Response = slicerouter_updated
 			return r.Status().Update(ctx, workerslicegwrecycler)
-		})
+		}, retry.Attempts(20))
```
YachikaRalhan, line 274: Why retry number is comparatively less here? If it is expected, add a comment here.

PR #278, opened 2023-07-25
fix(): added finalizer to slicegw (handled cluster detach scenario)

YachikaRalhan:
added finalizer to slicegw for handling cluster detach scenario

### DIFF
--- a/pkg/hub/controllers/slicegateway_controller.go
+++ b/pkg/hub/controllers/slicegateway_controller.go
@@ -50,7 +50,7 @@ type SliceGwReconciler struct {
 }
 
 var sliceGWController = "workersliceGWController"
-var sliceGwFinalizer = "controller.kubeslice.io/hubSpokeSliceGw-finalizer"
+var sliceGwFinalizer = "controller.kubeslice.io/sliceGw-finalizer"
 
 func (r *SliceGwReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	log := logger.FromContext(ctx)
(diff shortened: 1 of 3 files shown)


pkg/hub/controllers/slicegateway_controller.go
```diff
@@ -50,6 +50,7 @@ type SliceGwReconciler struct {
 }
 
 var sliceGWController = "workersliceGWController"
+var sliceGwFinalizer = "controller.kubeslice.io/hubSpokeSliceGw-finalizer"
```
rahulsawra98, line 53: worker.kubeslice.io/sliceGw-finalizer

pkg/hub/controllers/slicegateway_controller.go
```diff
@@ -73,8 +74,12 @@ func (r *SliceGwReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		log.Info("sliceGw doesn't belong to this cluster", "sliceGw", sliceGw.Name, "cluster", clusterName, "slicegw cluster", sliceGw.Spec.LocalGatewayConfig.ClusterName)
 		return reconcile.Result{}, nil
 	}
+	requeue, result, err := r.handleSliceGWDeletion(sliceGw, ctx, req)
```
rahulsawra98, line 77: where are you adding this finalizer?
if you are not adding how will you get an event? 
PTAL whenever you get time

rahulsawra98 (changes requested):

pkg/hub/controllers/slicegateway_controller.go
YachikaRalhan: Adding it inside this function on line number 213

pkg/hub/controllers/slicegateway_controller.go
YachikaRalhan: I am following the same naming convention like slice controller at hub side has.
I can remove the hubSpoke after slash but the prefix should be controller IMO to make it consistent

rajendra-avesha (approved):

PR #281, opened 2023-07-31
fix(): wait for gw pods to be in ready state before marking complete

YachikaRalhan:


pkg/hub/controllers/vpnkeyrotation/reconciler.go
```diff
@@ -380,9 +380,15 @@ func (r *Reconciler) syncCurrentRotationState(ctx context.Context,
 		syncedRotationState := make(map[string]hubv1alpha1.StatusOfKeyRotation)
 		for _, gw := range allGwsUnderCluster {
 			obj, ok := currentRotationState[gw]
+
 			if ok {
 				syncedRotationState[gw] = obj
 			} else if !ok {
+				gwIsReady := r.areGatewayPodsReady(ctx, gw)
+				if !gwIsReady {
```
rahulii, line 388: lets just use
if !r.areGatewayPodsReady(ctx, gw)

rahulii:
Lets add UTs and ITs if posibble.

pkg/hub/controllers/vpnkeyrotation/reconciler.go
```diff
@@ -366,10 +366,10 @@ func (r *Reconciler) syncCurrentRotationState(ctx context.Context,
 			return getErr
```
rajendra-avesha, line 366: @YachikaRalhan this method returns (bool, error) you are returning one value

pkg/hub/controllers/vpnkeyrotation/reconciler.go
```diff
@@ -380,9 +380,15 @@ func (r *Reconciler) syncCurrentRotationState(ctx context.Context,
...
 		for _, gw := range allGwsUnderCluster {
 			obj, ok := currentRotationState[gw]
+
 			if ok {
 				syncedRotationState[gw] = obj
 			} else if !ok {
+				gwIsReady := r.areGatewayPodsReady(ctx, gw)
+				if !gwIsReady {
+					requeue = true
+					return errors.New("gateway pods are not ready")
```
rajendra-avesha, line 390: @YachikaRalhan you need to requeue also here (bool, error)

pkg/hub/controllers/vpnkeyrotation/reconciler.go
```diff
@@ -474,6 +480,25 @@ func (r *Reconciler) updateCertificates(ctx context.Context, rotationVersion int
...
+		"kubeslice.io/slice-gw": sliceGWName,
+	}
+	listOptions := []client.ListOption{
+		client.MatchingLabels(labels),
+	}
+	err := r.WorkerClient.List(ctx, &podsList, listOptions...)
+	if err != nil {
+		return false
+	}
+	if len(podsList.Items) == 2 {
```
rajendra-avesha, line 496: @YachikaRalhan will the pod count be always 2?

rajendra-avesha (changes requested):

pkg/hub/controllers/vpnkeyrotation/reconciler.go
YachikaRalhan: This is under errors.Retry function which is returning only bool.

pkg/hub/controllers/vpnkeyrotation/reconciler.go
YachikaRalhan: Yes each gateway have two deployments/pods

pkg/hub/controllers/vpnkeyrotation/reconciler.go
YachikaRalhan: @rajendra-avesha that is under errors.Retry which has a callback function returning error only

rajendra-avesha (approved):

PR #333, opened 2024-02-15
fix(): Reconciles port numbers in the ovpn client deployment

### LINKED ISSUE #309: Reconcile port numbers in ovpn client deployments
### 📜 Description

An ovpn client deployment for a slice gateway is created with the NodePort port numbers specified in the workerslicegateway object. The logic to handle a change in the port numbers seems to be missing. The ovpn clients continue to connect on the previously specified port numbers. The ovpn client deployment needs to be updated to effect a change in the port number. 

### 👟 Reproduction steps

There are two options:
1. Delete the slice object manually on the designated server cluster. This will lead to deletion of the slicegateway objects as well. Make sure the NodePort services are also deleted. Then restart the worker-operator pod. This will lead to re-creation of the slice and slicegateway objects, and new NodePort services. The workerslicegateway objects are updated with new port numbers, but on the client cluster the port number is not updated in the ovpn deployment.

2. Manually edit the client deployment and update the port number to a random value. The worker-operator does not reconcile the deployment back to valid port numbers.  

### 👍 Expected behavior

The worker-operator on the client cluster should reconcile the ovpn deployments to use currently relevant port numbers.

### 👎 Actual Behavior

The ovpn client deployments continue to use invalid port numbers leading to loss in connectivity between clusters.

### 🐚 Relevant log output

_No response_

### Version

_No response_

_No response_

### ✅ Proposed Solution

_No response_

- [X] I checked and didn't find any similar issue

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kon3m:
# Description

Fixes #309 

## How Has This Been Tested?
Tested it manually using two single node kind clusters, will write test cases soon using envtest and update the PR

## Checklist:

* [x] I have ran `go fmt`
* [ ] I have performed a self-review of my own code.
* [ ] I have tested it for all user roles.

```release-note

```


narmidm:
image

one issue notices the commits are unverified (No user is associated with the committer email). We require every contributor to certify that they are legally permitted to contribute to our project. A contributor expresses this by consciously signing their commits,

please follow this link
https://github.com/kubeslice/worker-operator/blob/master/CONTRIBUTING.md#contributor-compliance-with-developer-certificate-of-origin-dco

controllers/slicegateway/slicegateway.go
```diff
@@ -1534,3 +1518,30 @@ func (r *SliceGwReconciler) ReconcileIntermediateGatewayDeployments(ctx context.
...
+
+// updateGatewayDeployment updates the gateway client deployments with the relevant updated ports
+// from the workersliceconfig
+func (r *SliceGwReconciler) updateGatewayDeployment(ctx context.Context, c client.Client, g *kubeslicev1beta1.SliceGateway, deployment *appsv1.Deployment, nodePort int) error {
+	containers := deployment.Spec.Template.Spec.Containers
+	for contIndex, cont := range containers {
+		if cont.Name == "kubeslice-sidecar" {
+			for index, key := range cont.Env {
+				if key.Name == "NODE_PORT" {
+					upEnvVar := corev1.EnvVar{
```
Rahul-D78, line 1530: Can we make this var name as `updatedEvnVar` instead of `upEnvVar`

controllers/slicegateway/slicegateway.go
```diff
@@ -1534,3 +1518,30 @@ func (r *SliceGwReconciler) ReconcileIntermediateGatewayDeployments(ctx context.
 
 	return ctrl.Result{}, nil, false
 }
+
+// updateGatewayDeployment updates the gateway client deployments with the relevant updated ports
+// from the workersliceconfig
+func (r *SliceGwReconciler) updateGatewayDeployment(ctx context.Context, c client.Client, g *kubeslicev1beta1.SliceGateway, deployment *appsv1.Deployment, nodePort int) error {
```
Rahul-D78, line 1524: Can we make this function name as `updateGatewayDeploymentNodePort` because this function is only updating nodePort info in the deployment.

Rahul-D78 (commented):
Looks great! Thanks for adding this and also for fixing the flaky test  :+1:

Rahul-D78 (approved):
LGTM, thank you @kon3m

kon3m:
Thanks for reviewing @Rahul-D78

narmidm:
@kon3m can we put PR to ready to review.

kon3m:
@narmidm i have found an issue with this fix where in some cases distinct ports do not get assigned to the client deployments or to put in another way both the clients are connected to the same nodePort which should not happen. Thats the reason i have converted the PR to draft. Now i have updated the PR adding the fix for that as well, could you please review it?

narmidm (approved):

narmidm:
@kon3m , can you rebase with the latest master. it has some fixes of E2E pipeline. After That I will trigger the E2E again.

kon3m:
@narmidm done, i have just rebased with master and pushed.

controllers/slicegateway/slicegateway.go
```diff
@@ -1377,16 +1358,32 @@ func (r *SliceGwReconciler) ReconcileGatewayDeployments(ctx context.Context, sli
 
 	// Reconcile deployment to node port mapping for gw client deployments
 	if isClient(sliceGw) {
+		UpdateDeploymentSlice := make([]appsv1.Deployment, 0)
+		toUpdate := false
 		for _, deployment := range deployments.Items {
 			found, nodePortInUse := getClientGwRemotePortInUse(ctx, r.Client, sliceGw, deployment.Name)
 			if found {
```
bharath-avesha, line 1365: I think this block can be simplified and made more readable if we first check if the port in use is actually present in the list of remote port numbers in the slicegw object. 

```
for _, deployment := range deployments.Items {
			found, nodePortInUse := getClientGwRemotePortInUse(ctx, r.Client, sliceGw, deployment.Name)
			if found {
                            // Check if the portInUse is valid. 
                            // It is valid only if the list of remoteNodePorts in the slicegw object contains the portInUse.
                            if !checkIfNodePortIsValid(sliceGw.Status.Config.SliceGatewayRemoteNodePorts, 
                                                                       nodePortInUse) {
                                // Get a valid port number for this deployment
                                portNumToUpdate := 
                                    allocateNodePortToClient(sliceGw.Status.Config.SliceGatewayRemoteNodePorts,     
                                                                                &gwClientToRemotePortMap)
                                // Update the port map
                                gwClientToRemotePortMap.Store(deployment.Name, portNumToUpdate)
                                err := r.updateGatewayDeploymentNodePort(ctx, r.Client, sliceGw, &deployment, portNumToUpdate)
                                if err != nil {
                                    return ctrl.Result{}, err, true
                                }
                                // Requeue if the update was successful
                                return ctrl.Result{}, nil, true
                          }
                          // At this point, the node port in use is valid. Check if the port map is populated. Only case
                         // where it is not populated is if the operator restarts. The populated value must match the
                         // port in use. If not, the deploy needs to be updated to match the state stored in the operator.
                         portInMap, foundInMap := gwClientToRemotePortMap.Load(deployment.Name) 
                         if foundInMap {
                             if portInMap != nodePortInUse {
                                 // Update the deployment since the port numbers do not match
                                 err := r.updateGatewayDeploymentNodePort(ctx, r.Client, sliceGw, &deployment, portInMap)
                                if err != nil {
                                    return ctrl.Result{}, err, true
                                }
                                // Requeue if the update was successful
                                return ctrl.Result{}, nil, true
                         } else {
                             gwClientToRemotePortMap.Store(deployment.Name, nodePortInUse)
                         }
               }
}
```

Note: checkIfNodePortIsValid() is already defined in the utils.go file.

controllers/slicegateway/slicegateway.go
```diff
@@ -42,19 +50,10 @@ import (
...
-	"github.com/kubeslice/worker-operator/pkg/cluster"
-	"github.com/kubeslice/worker-operator/pkg/gwsidecar"
-	"github.com/kubeslice/worker-operator/pkg/logger"
-	"github.com/kubeslice/worker-operator/pkg/router"
-	"github.com/kubeslice/worker-operator/pkg/utils"
-	webhook "github.com/kubeslice/worker-operator/pkg/webhook/pod"
 )
 
 var (
+	certFileName             = "openvpn_client.ovpn"
```
bharath-avesha, line 56: Can we plz change this to vpnClientFileName?

controllers/slicegateway/slicegateway.go
kon3m: @bharath-avesha thanks for the review. I initially went with the same approach but there was a weird case where the nodePorts in the slicegw config have same values (this happens only once or twice after reconciler recognises the change in nodeports from the controller side). When this happens the the values in the map are altered (both the deployments having the same values). So for this reason in the first loop i am making sure that the map has correct values and then updating the deployments(if any) in the second. I will try to reproduce the issue and update the PR accordingly.

bharath-avesha: Thanks for the update @kon3m.. Lets try this change and see if the duplicate port number issue comes up.. Plz collect the operator logs and let me know..

kon3m: @bharath-avesha i have tried to reproduce it yesterday and it did show up again and i wanted to record/screencast the scenario but it is not consistent and was unable to capture/reproduce it again. Can i just apply the above change with a line before the loop like:

`if hasSameValues(sliceGw.Status.Config.SliceGatewayRemoteNodePorts) {
    return ctrl.Result{}, nil, true
 }
` 
just requeue if it has same port numbers.

kon3m: @bharath-avesha i have tried reproducing the scenario after applying the above change , the issue doesn't seem to show up. Updated the PR with the change.

controllers/slicegateway/slicegateway.go
```diff
@@ -1392,13 +1373,40 @@ func (r *SliceGwReconciler) ReconcileGatewayDeployments(ctx context.Context, sli
 		for _, deployment := range deployments.Items {
 			found, nodePortInUse := getClientGwRemotePortInUse(ctx, r.Client, sliceGw, deployment.Name)
 			if found {
-				_, foundInMap := gwClientToRemotePortMap.Load(deployment.Name)
-				if !foundInMap {
+				// Check if the portInUse is valid.
+				// It is valid only if the list of remoteNodePorts in the slicegw object contains the portInUse.
+				if !checkIfNodePortIsValid(sliceGw.Status.Config.SliceGatewayRemoteNodePorts, nodePortInUse) {
+					// Get a valid port number for this deployment
+					portNumToUpdate := allocateNodePortToClient(sliceGw.Status.Config.SliceGatewayRemoteNodePorts, &gwClientToRemotePortMap)
```
bharath-avesha, line 1380: The allocateNodePort func seems to return a port number not already stored in the port map, and returns 0 if the input list of ports and the ports stored in the map match. What if the port map already had an entry for the deployment? It would end up returning 0, right? So i think the alloc func needs to also take the name of the deployment as input. If it cannot find a free port in the list, and before returning a 0, it should check if the map has an entry for the deployment. If yes, return the value of the entry, else return 0.

Please also check for the return value. If 0, return an error.

kon3m: Great find! i actually was handling this case before (checking if its zero in the caller) and forgot to update it in the recent change i made . Will update the PR soon and let you know.

kon3m: @bharath-avesha i have updated the PR, please have a look.

bharath-avesha (approved):

kon3m:
@narmidm could you please trigger the E2E pipeline?

NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2024-03-21T08:01:56-master-444/index.html'

NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2024-03-27T07:14:18-master-445/index.html'

NishantSingh10:
report link  'https://kubeslice.github.io/e2e-allure-reports/Kind-worker-operator-2024-03-29T13:01:53-master-446/index.html'

kon3m:
Opened a new clean [PR]( https://github.com/kubeslice/worker-operator/pull/368)

PR #408, opened 2024-10-29
fix(): configurable ns exclusion list

priyank-upadhyay:
# Description
This provides a configurable way to provide a list of namespaces to be excluded in namespace reconcilers. It can be passed through an environment variable. If not provided a default list would be used.

Fixes # 

## Checklist:

* [x] I have ran `go fmt`
* [x] I have tested it for all user roles.

```release-note

```

### DIFF
--- a/pkg/namespace/controllers/reconciler.go
+++ b/pkg/namespace/controllers/reconciler.go
@@ -19,6 +19,7 @@ package namespace
 
 import (
 	"context"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/kubeslice/kubeslice-monitoring/pkg/events"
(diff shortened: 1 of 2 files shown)


narmidm (approved):

pkg/namespace/controllers/reconciler.go
```diff
@@ -68,6 +68,9 @@ func (c *Reconciler) getSliceNameFromNs(ns string) (string, error) {
 }
 
 func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	excludedNsEnv := utils.GetEnvOrDefault("EXCLUDED_NS", "kube-system,default,kubeslice-system,kube-node-lease,kube-public,istio-system")
```
bharath-avesha, line 72: Can we have a global var defaultExcludedNS with all the namespaces we want to exclude? then we could do: utils.GetEnvOrDefault("EXCLUDED_NS", strings.Join(defaultExcludedNs, ","))

priyank-upadhyay: made the change.
strings.Join and split would be unnecessary so just using comma separated string in the global const.

bharath-avesha (approved):

PR #416, opened 2025-01-16
fix(): Avoid unnecessary reconcile cycles for slice gw

bharath-avesha:
The slice gw reconciler is in a perpetual reconcile cycle. There are places where the CR update func is being called in a for loop. And reflect.DeepEqual is being used to compare two lists that are never guaranteed to have their elements in the same order, this causes unnecessary update from the server worker cluster to the controller, which triggers an update on the client cluster. The reconciler also watches for node updates blindly, triggering the reconcile func every time there is an update to the Node object. The update could be because of the cluster provider adding or modifying labels and annotations. We need a granular check on the node update and trigger the reconciler func only if needed.

Due to the reconcile storm, updates are not being handled in a timely manner, and sometimes it is so bad that the reconcile events get queued for days. This was observed on a customer cluster where an update to the CR was not handled for 15 days.

### DIFF
--- a/controllers/slicegateway/reconciler.go
+++ b/controllers/slicegateway/reconciler.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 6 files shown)


gourishkb (approved):

mridulgain (approved):

Rahul-D78 (approved):

Rahul-D78:
Looks Good to me, we can also try to always make sure what ever updates we are doing to the CR is always `sorted`. so if we are doing checks further we can avoid this issue of DeepEqual.

PR #446, opened 2025-08-19
fix(): Optimize NetworkPolicy reconciler to reduce unnecessary API update calls and reduce log noise

### LINKED ISSUE #372: Bug: Refactor NetworkPolicy reconciler
### 📜 Description

Currently, the NetworkPolicy reconciler is updating the NetworkPolicy resource in the application namespace even if there is no change. It also updates the `slice.Status.NetworkPoliciesInstalled` field to true during each reconciliation interval. And generating events and logs in the for loop.

### 👟 Reproduction steps

Create a slice resource and onboard few application namespaces. You can see logs like `Installed netpol for namespace successfully` and `Updated network policy` very frequently.

### 👍 Expected behavior

It should update the `NetworkPolicy` and `Slice` resource conditionally when update is required. And generate the logs and events after the create / update rather than generating inside a for loop.

### 👎 Actual Behavior

Currently the reconciler is performing the update calls to the k8s api server very frequently. Which might impact the performance if the number of resources grows.

### 🐚 Relevant log output

_No response_

### Version

_No response_

_No response_

### ✅ Proposed Solution

The below line can be simlified by checking if the `NetworkPoliciesInstalled` field is false then only set it to true.

Before:
```GO
slice.Status.NetworkPoliciesInstalled = true
return r.Status().Update(ctx, slice)
```
After:
```GO
if !slice.Status.NetworkPoliciesInstalled {
	slice.Status.NetworkPoliciesInstalled = true
	return r.Status().Update(ctx, slice)
}
return nil
``` 

Instead of generating log in a for we can generate it after the netpol resource got created for the first time, Currently It is generating logs and event for each iteration for both create and update.
```GO
for _, appNsObj := range appNsList.Items {
	err = r.installSliceNetworkPolicyInAppNs(ctx, slice, appNsObj.ObjectMeta.Name)
	if err != nil {
           ....
        }
	utils.RecordEvent(ctx, r.EventRecorder, slice, nil, ossEvents.EventNetPolAdded, "slice_reconciler")
	log.Info("Installed netpol for namespace successfully", "namespace", appNsObj.ObjectMeta.Name)
}
```

In the `installSliceNetworkPolicyInAppNs` method we are updating the resource in each reconciliation interval, Instead we can get the actual resource if it not found then we can create it else we can compare it with the constructed resource and update it if it not equal. This line `log.Info("Updated network policy", "namespace", appNs)` can be called when there is an update.

- [X] I checked and didn't find any similar issue

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

greninja517:
# Description

This PR optimizes the NetworkPolicy reconciliation logic with following changes:
- **Conditional NetworkPolicy updates**: Use `equality.Semantic.DeepEqual()` to compare specs and only update when different
- **Reduced logging frequency**: Move logs and events outside the reconciliation loop
- **Smart status updates**: Only update `NetworkPoliciesInstalled` when the field value actually changes
- **Better error handling**: Continue processing all namespaces even if some fail

#### No breaking changes - all existing functionality preserved with improved performance

Fixes #372  

## Checklist:

* [x] I have ran `go fmt`
* [x] I have tested it for all user roles.

```release-note

```

### DIFF
--- a/controllers/slice/namespaces.go
+++ b/controllers/slice/namespaces.go
@@ -696,7 +696,9 @@ func (r *SliceReconciler) reconcileSliceNetworkPolicy(ctx context.Context, slice
 	policiesInstalled := successfulNamespaces > 0
 	if slice.Status.NetworkPoliciesInstalled != policiesInstalled {
 		slice.Status.NetworkPoliciesInstalled = policiesInstalled
-		return r.Status().Update(ctx, slice)
+		if err := r.Status().Update(ctx, slice); err != nil {
+			return err
+		}
 	}
 
 	if len(errors) > 0 {


controllers/slice/namespaces.go
```diff
@@ -668,33 +669,71 @@ func (r *SliceReconciler) reconcileSliceNetworkPolicy(ctx context.Context, slice
...
+
+	policiesInstalled := successfulNamespaces > 0
+	if slice.Status.NetworkPoliciesInstalled != policiesInstalled {
+		slice.Status.NetworkPoliciesInstalled = policiesInstalled
+		if err := r.Status().Update(ctx, slice); err != nil {
+			return err
+		}
+	}
+
+	if len(errors) > 0 {
```
gourishkb, line 704: with this change we return an error after we have set the NetworkPoliciesInstalled status to true above even when only a part of the list of namespaces have the policies installed. 

I would suggest to keep the current idea of keeping the NetworkPoliciesInstalled status to be false even when there is partial failure. Otherwise this gives the wrong impression to a regular user that all namespaces are network isolated according to the plan.

greninja517: @gourishkb , I was about to implement the same but here https://github.com/kubeslice/worker-operator/blob/master/api/v1beta1/slice_types.go#L131-L133, it is mentioned that  `NetworkPoliciesInstalled` signifies at least one namespace has successfully `NetworkPolicies` installed. So, I have implemented the above approach.

PR #475, opened 2026-05-12
fix: return nil explicitly in SliceGateway NONET early-exit path

### LINKED ISSUE #474: Bug: SliceGateway reconciler returns `err` (which is nil) for NONET slices, masking the intent
### 📜 Description

`controllers/slicegateway/reconciler.go` lines 134–138:

```go
if slice.Status.SliceConfig != nil &&
    slice.Status.SliceConfig.SliceOverlayNetworkDeploymentMode == v1alpha1.NONET {
    log.Info("No communication slice. Skipping slicegw reconcilation")
    return ctrl.Result{}, err  // err is nil here but variable reuse is confusing
}
```

`err` at this point holds the result from `controllers.GetSlice` (line 122), which is `nil` if the slice was found. While this works correctly, using `err` in the return obscures that this is an intentional early exit with no error. Future readers may not trace back to verify `err` is nil, and a refactor that moves this block could silently break.

### 👟 Reproduction steps

Read `controllers/slicegateway/reconciler.go` lines 122–138.

### 👍 Expected behavior

`return ctrl.Result{}, nil` to make the intent explicit.

### 👎 Actual Behavior

`return ctrl.Result{}, err` where `err` happens to be nil but is not obviously so.

### 🐚 Relevant log output

```shell

```

### Version

_No response_

_No response_

### ✅ Proposed Solution

Replace `return ctrl.Result{}, err` with `return ctrl.Result{}, nil`.

- [x] I checked and didn't find any similar issue

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

mdryaan:
# Description

In `controllers/slicegateway/reconciler.go`, the early-exit path for NONET slices returns `err` instead of `nil`:

```go
// before
return ctrl.Result{}, err

// after  
return ctrl.Result{}, nil
```

At this point in the code, `err` holds the return value from `controllers.GetSlice` which was already checked to be non-nil a few lines above — so it is always `nil` here and behaviour is identical. The problem is that using `err` makes the intent unclear and creates a latent risk: any future refactor that introduces a new `err` assignment between the `GetSlice` call and this NONET guard could silently turn a clean early exit into an unexpected error return.

Fixes #474 

## How Has This Been Tested?

- [x] `make fmt` — no formatting changes
- [x] `make vet` — no issues
- [x] `make build` — builds cleanly

## Checklist:

* [x] I have ran `go fmt`
* [ ] I have commented my code, particularly in hard-to-understand areas.
* [ ] I have tested it for all user roles.

```release-note

```

### DIFF
--- a/controllers/slicegateway/reconciler.go
+++ b/controllers/slicegateway/reconciler.go
@@ -134,7 +134,7 @@ func (r *SliceGwReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	if slice.Status.SliceConfig != nil &&
 		slice.Status.SliceConfig.SliceOverlayNetworkDeploymentMode == v1alpha1.NONET {
 		log.Info("No communication slice. Skipping slicegw reconcilation")
-		return ctrl.Result{}, err
+		return ctrl.Result{}, nil
 	}
 
 	// Check if slice router network service endpoint (NSE) is present before spawning slice gateway pod.


mdryaaan:
Hey @gourishkb @richiesebastian @Rahul-D78,  Could you please have a look at this PR whenever you have time? Thanks!

